PR #7, opened 2021-11-05
Add Electrical Component docstrings

AayushSabharwal:
Working on adding some docstrings, trying to follow the pattern in MTK.jl docs.

~Some questions:~
- ~Do the `ModelingToolkit.connect` methods need docstrings? Consequently, does `ElectricalPin` need one?~
- ~Do the ideal components need to specify how their variables are calculated?~
- ~I've studied OpAmps a bit, but I'm not particularly sure how this one works~
- ~How should the pins of sensors (and components in general) be documented? All two-pin components have pins `p` and `n`, so should this be mentioned in every sensor/component or left as an assumed standard? What about components with more pins?~

EDIT: I'm not sure why I didn't just take the docs from the README


AayushSabharwal:
It seems like documenting macro usages is not straightforward. The docstrings for `@connector function Pin` and `@connector function DigitalPin` seem to be the issue. I did come across [this](https://discourse.julialang.org/t/can-i-make-docstrings-work-with-everywhere/11402) but it doesn't seem to work here, just leads to another error in MTK.jl

AayushSabharwal:
Apologies for the inactivity. I have a lot going on right now. Would love to get back to this in a week!

AayushSabharwal:
Documenting `Pin` and `DigitalPin` requires `@doc` to be used, since they are the result of macro evaluations

AayushSabharwal:
If #18 is on its way to completion, it might be better to wait and migrate the docstrings over to those new functions? It would reduce a lot of duplicacy

AayushSabharwal:
I'm not familiar with `PulseDiff`, but the rest of the docstrings are done. I left out the current sources because of the above comment. They would be nearly identical to the docstrings of the corresponding voltage sources.

ChrisRackauckas:
Looks good thanks!

ChrisRackauckas:
Test failure

AayushSabharwal:
I'll look into this. Weird that docs failed tests.

AayushSabharwal:
It seems that `@register _and(x...)` is the issue. Symbolics deprecated `@register` for `@register_symbolic`, and it doesn't like that.

ChrisRackauckas:
we should just drop the earlier Symbolics and update.

AayushSabharwal:
I'm afraid this still won't precompile, because Symbolics doesn't like the `x...` in `@register_symbolic _and(x...)`

AayushSabharwal:
It seems the deprecation of `@register` for `@register_symbolic` broke this. If I use `@register` everywhere on Symbolics v4.1.0 this works fine

AayushSabharwal:
MTK v8 removed `promote_connect_rule` in https://github.com/SciML/ModelingToolkit.jl/pull/1348. What is the new API for this functionality?

ChrisRackauckas:
If the variables are appropriately defined as flow variables this should all happen automatically through the `connect` function.

AayushSabharwal:
I guess that should happen in #25 then. Symbolics v4.1.1 still breaks this

AayushSabharwal:
Tests should pass now with Symbolics v4.2.2 release

ChrisRackauckas:
Please update this to match master. We just had a pretty major change (https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/7) and it would be good to copy these over if possible.

ChrisRackauckas:
Looking at http://mtkstdlib.sciml.ai/dev/API/electrical/, I like how your format defines the ports and everything, so updating it would be great.

AayushSabharwal:
Sure! I'm a bit swamped with work right now, but I'll get to this as soon as I can.

ChrisRackauckas:
Oh no, don't merge, rebase this one 😓. You might want to just start from scratch and copy over the major changes.

AayushSabharwal:
I guess that's a better idea. I spent the past half hour trying to reconcile all of this only for it to show up again :P

Rebasing was annoying, since it seemed to process commit by commit and kept getting merge conflicts in the same files at the same places repeatedly, so I tried `--no-rebase`. Looks like git still doesn't like how this branch looks. I'll make another one and PR again.

PR #15, opened 2021-11-11
remove D equations from PID if D term turned off

baggepinnen:
The nl solver errored if the terms were included with coefficient 0

This PR also introduces a safeguard for the default value of the anti-windup tracking coefficient.


baggepinnen:
The failure seems unrelated to the PR; but the tests touching the PR never got run. Is it the SafeTests package that causes tests to quit early if something fails?

ChrisRackauckas:
Is it due to the version bump?

It's just a standard stdlib Test thing to stop at first failure.

baggepinnen:
> Is it due to the version bump?

I'm not sure what's causing this, it happens in alias elimitnation during the tests of analog components
https://github.com/SciML/ModelingToolkitStandardLibrary.jl/runs/4175646983?check_suite_focus=true#step:6:259

Perhaps @YingboMa has a clue?

PR #18, opened 2021-11-18
Decouples Signal sources from their smooth wave shapes

ven-k:
- Adds `wave_functions` to Blocks. The `*Function` components give out `y` output of `*` shape
- Tests the wave-functions
- Removes all shaped voltage sources in favour of `VoltageSource`

TODO:
- [ ] Add tests for CurrentSource + Functions

### DIFF
--- a/src/Blocks/wave_functions.jl
+++ b/src/Blocks/wave_functions.jl
@@ -1,0 +1,131 @@
+# Define and register smooth functions
+using ModelingToolkit, Symbolics
+_constant(K) = K
+_cos_wave(x, f, A, st, ϕ) = A*cos(2*π*f*(x-st) + ϕ)
+_damped_sine_wave(x, f, A, st, ϕ, d) = exp((st-x)*d)*A*sin(2*π*f*(x-st) + ϕ)
+_ramp(x, δ, st, et, h) = h/(et-st)*(_xH(x, δ, st) - _xH(x, δ, et))
+_sine(x, f, A, st, ϕ) = A*sin(2*π*f*(x - st) + ϕ)
+_square_wave(x, δ, f, A, st) = A*2atan(sin(2π*(x-st)*f)/δ)/π
+_step(x, δ, h, a) = h*(atan((x-a)/δ)/π + 0.5)
+_triangular_wave(x, δ, f, A, st) = A*(1-2acos((1 - δ)sin(2π*(x-st)*f))/π)
+_xH(x, δ, tₒ) = 0.5*(x-tₒ)*(1+((x-tₒ)/sqrt((x-tₒ)^2+δ^2)))
+
+
+@register _constant(K)
+@register _cos_wave(x, f, A, st, ϕ)
+@register _damped_sine_wave(x, f, A, st, ϕ, damping)
+@register _ramp(x, δ, st, et, h)
+@register _sine(x, f, A, st, ϕ)
+@register _square_wave(x, δ, f, A, st)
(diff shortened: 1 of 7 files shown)


src/Blocks/wave_functions.jl
```diff
@@ -0,0 +1,131 @@
+# Define and register smooth functions
+using ModelingToolkit, Symbolics
+_constant(K) = K
+_cos_wave(x, f, A, st, ϕ) = A*cos(2*π*f*(x-st) + ϕ)
+_damped_sine_wave(x, f, A, st, ϕ, d) = exp((st-x)*d)*A*sin(2*π*f*(x-st) + ϕ)
+_ramp(x, δ, st, et, h) = h/(et-st)*(_xH(x, δ, st) - _xH(x, δ, et))
+_sine(x, f, A, st, ϕ) = A*sin(2*π*f*(x - st) + ϕ)
+_square_wave(x, δ, f, A, st) = A*2atan(sin(2π*(x-st)*f)/δ)/π
+_step(x, δ, h, a) = h*(atan((x-a)/δ)/π + 0.5)
```
baggepinnen, line 9:
```suggestion
_step(x, δ, h, a) = h*(atan((x-a)/δ)/π + 1//2)
```
something like this would perhaps be nice to not auto promote to `Float64`

src/Blocks/wave_functions.jl
```diff
@@ -0,0 +1,131 @@
...
+_triangular_wave(x, δ, f, A, st) = A*(1-2acos((1 - δ)sin(2π*(x-st)*f))/π)
+_xH(x, δ, tₒ) = 0.5*(x-tₒ)*(1+((x-tₒ)/sqrt((x-tₒ)^2+δ^2)))
```
baggepinnen, line 11:
```suggestion
_xH(x, δ, tₒ) = (x-tₒ)*(1+((x-tₒ)/sqrt((x-tₒ)^2+δ^2)))/2
```

ChrisRackauckas:
I'm not sure this applies after https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/7.

ven-k:
#7 wouldn't.

This PR moves smooth functions to `Blocks` and then exports only one `VoltageSource` and a `CurrentSource`. Mix and match of smooth wave generators in Blocks and these sources will give out the signals of required shapes.
This reduces `n_shapes`*`n_dimensions` to `n_shapes`+ `n_dimensions`.

For example
(provided by @baggepinnen)
image
where a sin source is connected to a torque converter. The torque converter could easily be replaced for a voltage/flow/current converter for a different application

Current status:
- [x] Signals and Shapes are already decoupled
- [ ]  Doesn't work for non zero starttime https://github.com/SciML/ModelingToolkit.jl/issues/1409

(If necessary I can set all starttime to 0; and rebase to `main` . The unit tests will pass)

ValentinKaisermayer:
> I'm not sure this applies after #7.

I guess you meant #26

ValentinKaisermayer:
A similar example to the one above is already present in main. https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/99b273da40fa7418c78d6335a3ef7c752d2ece18/test/Mechanical/rotational.jl#L31

What would be needed are corresponding smooth source functions in `Blocks`, see #30.

ven-k:
> What would be needed are corresponding smooth source
Perfect. This adds smooth functions.

PR #24, opened 2021-12-23
Revert "CompatHelper: bump compat for ModelingToolkit to 8, (keep existing compat)"

ChrisRackauckas:
Reverts SciML/ModelingToolkitStandardLibrary.jl#23


ChrisRackauckas:
Should not have merged that with the others 😓

PR #27, opened 2022-04-11
Fix v8 connector

ValentinKaisermayer:
Just removes the custom connect methods and adds defaults.


ValentinKaisermayer:
I do not think this will be enough to pass the tests.

Issue #28, opened 2022-04-20
Rotational mechanical components


ValentinKaisermayer:
# Components

- [x] Fixed
- [x] Inertia
- [x] Spring
- [x] Damper
- [x] IdealGear

# Sources
- [x] Torque

# Sensors (abs. and rel.)
- [x] Angle
- [x] Speed
- [x] Torque

PR #34, opened 2022-04-23
Adds rotational components

ValentinKaisermayer:
see #28

### DIFF
--- a/test/Mechanical/rotational.jl
+++ b/test/Mechanical/rotational.jl
@@ -1,0 +1,101 @@
+using ModelingToolkitStandardLibrary.Mechanical.Rotational, ModelingToolkit, OrdinaryDiffEq, Test
+import ModelingToolkitStandardLibrary.Blocks
+# using Plots
+
+@parameters t
+
+@testset "two inertias" begin
+    @named fixed = Fixed()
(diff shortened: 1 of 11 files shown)


test/Mechanical/rotational.jl
```diff
@@ -0,0 +1,101 @@
...
+        connect(damper.flange_b, fixed.flange)
+        connect(sine.output, torque.tau)
+        connect(torque.support, fixed.flange)
+        connect(idealGear.support, fixed.flange)
+        connect(torque.flange, inertia1.flange_a)
+    ]
+
+    @named model = ODESystem(connections, t, systems=[fixed, torque, inertia1, idealGear, inertia2, spring, inertia3, damper, sine])
+    sys = structural_simplify(model)
+    @test_broken prob = ODAEProblem(sys, Pair[], (0, 1.0)) # KeyError: key 25 not found
```
ValentinKaisermayer, line 97: @ChrisRackauckas Any ideas why I get this error?

ChrisRackauckas: Open an issue on this to follow up. This seems like an @YingboMa question, but it's a good MWE.

PR #39, opened 2022-04-25
extracts _clamp and _dead_zone and makes tests more strict

ValentinKaisermayer:


ChrisRackauckas:
Yeah, it would probably be good to declare those for the discontinuities, and have a kwarg for not including them.

PR #41, opened 2022-04-28
Electrical Component Docstrings

AayushSabharwal:
Copied over from #7 without the git issues.


ChrisRackauckas:
Can you also add the new docstrings to https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/main/docs/src/API/electrical.md ?

src/Electrical/Analog/ideal_components.jl
```diff
@@ -28,13 +47,27 @@ function Resistor(;name, R=1.0)
...
-    Capacitor(;name, C=1.0, v_start=0.0) 
+```julia
+function Capacitor(; name, C = 1.0)
+```
+
+Creates an ideal Capacitor.
+
+# States
+- `v(t)`: [`V`]
+  The voltage across the capacitor, given by `D(v) ~ p.i / C`
```
ValentinKaisermayer, line 58: The states are actually inherited from the `OnePort` component, hence they are `v` and `i`. Maybe a `@ref` would do it.

ValentinKaisermayer:
> Can you also add the new docstrings to https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/main/docs/src/API/electrical.md ?

The Digital stuff isn't tested and the include is commended out, so the doc build will fail I think.

AayushSabharwal:
Documenter doesn't fail when docstrings aren't defined, it throws a warning and renders an error message in the docs, which appears to be what's happening here judging by the CI logs.

ChrisRackauckas:
IMO, let's merge this and let it throw a warning in the docs for now. It will leave clear markers of what needs to be finished still, but let people know it's coming soon.

PR #44, opened 2022-04-30
Adds missing links in docs and moves plot to assets

ValentinKaisermayer:


ChrisRackauckas:
Remove the plot from assets and squash

ValentinKaisermayer:
> Remove the plot from assets and squash

Where to put it then? 

Currently there is this plot (mistake?): https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/main/plot.png

ChrisRackauckas:
It should just be pasted into the issues and used as a link.

Issue #29, opened 2022-04-20
DC Permanent Machine


ValentinKaisermayer:
For a first academic use and example

- [x] add EMF
- [x] add el. mech. converter
- [x] add friction
- [x] add tests
- [x] add doc

PR #45, opened 2022-04-30
DCPM

ValentinKaisermayer:
The goal is to be able to model a simple DC permanent machine with PI controller.


ValentinKaisermayer:
related to #29 and #28

PR #55, opened 2022-05-14
Adds meta data

ValentinKaisermayer:
MTK now throws a warning if a connector has no flow, stream or I/O variables

### DIFF
--- a/src/Blocks/utils.jl
+++ b/src/Blocks/utils.jl
@@ -1,8 +1,8 @@
 @connector function RealInput(;name, nin=1, u_start=nin > 1 ? 0.0 : zeros(nin))
     if nin == 1
-        @variables u(t) = u_start
+        @variables u(t) = u_start [input=true]
     else
-        @variables u[1:nin](t) = u_start
+        @variables u[1:nin](t) = u_start [input=true]
         u = collect(u)
     end
     ODESystem(Equation[], t, [u...], []; name=name)
@@ -22,9 +22,9 @@ Connector with one input signal of type Real.
 
 @connector function RealOutput(;name, nout=1, u_start=nout > 1 ? 0.0 : zeros(nout))
     if nout == 1
-        @variables u(t) = u_start
+        @variables u(t) = u_start [output=true]
     else
-        @variables u[1:nout](t) = u_start
+        @variables u[1:nout](t) = u_start [output=true]
         u = collect(u)
     end
     ODESystem(Equation[], t, [u...], []; name=name)


ValentinKaisermayer:
@YingboMa there are some interesting new exceptions here, that are certainly not related to the changes in this PR or are they?

YingboMa:
It could be related. I recommend to not set flow variables to 0, since zero flow rate is often the singular case.

ValentinKaisermayer:
I got two times 
` During the resolution of the non-linear system, the evaluation of the following equation(s) resulted in a non-finite number`
for the tests of the `Blocks`, where the only change is, that I marked all inputs with `input=true` and outputs with `output=true`. Does that change anything for MTK?

ValentinKaisermayer:
> I got two times ` During the resolution of the non-linear system, the evaluation of the following equation(s) resulted in a non-finite number` for the tests of the `Blocks`, where the only change is, that I marked all inputs with `input=true` and outputs with `output=true`. Does that change anything for MTK?

Somehow it does change the behavior! The following example errors if I set the inputs and outputs meta data and runs fine if I do not:
```julia
@named c1 = Sine(; frequency=1)
@named c2 = Constant(; k=2)
@named div = Division(;)
@named int = Integrator(; k=1)
@named model = ODESystem(
    [
        connect(c1.output, div.input1), 
        connect(c2.output, div.input2), 
        connect(div.output, int.input),
    ], 
    t, 
    systems=[int, div, c1, c2]
)
prob = ODEProblem(sys, Pair[int.x=>0.0], (0.0, 1.0))
sol = solve(prob, Rodas4())
```
```julia
ERROR: During the resolution of the non-linear system, the evaluation of the following equation(s) resulted in a non-finite number: [3]
```

It seems MTK is not able to handle this system with the IO meta data. If I have a look at the equations of the simplified system they are horrible:
```julia
julia> equations(sys)
4-element Vector{Equation}:
 Differential(t)(int₊x(t)) ~ int₊k*int₊u(t)
 0 ~ div₊input1₊u(t) - c1₊output₊u(t)
 0 ~ div₊input2₊u(t) - c2₊output₊u(t)
 0 ~ int₊input₊u(t) - div₊output₊u(t)

julia> observed(sys)
9-element Vector{Equation}:
 int₊output₊u(t) ~ int₊x(t)
 int₊y(t) ~ int₊x(t)
 c2₊output₊u(t) ~ c2₊k
 c1₊output₊u(t) ~ c1₊offset + IfElse.ifelse(t < c1₊start_time, 0, c1₊amplitude*sin(c1₊phase + 6.283185307179586c1₊frequency*(t - c1₊start_time)))
 div₊input2₊u(t) ~ c2₊output₊u(t)
 div₊input1₊u(t) ~ c1₊output₊u(t)
 div₊output₊u(t) ~ c1₊output₊u(t) / c2₊output₊u(t)
 int₊u(t) ~ div₊output₊u(t)
 int₊input₊u(t) ~ div₊output₊u(t)
```
While for the simplified system without IO meta data:
```julia
julia> equations(sys)
1-element Vector{Equation}:
 Differential(t)(int₊x(t)) ~ int₊k*div₊output₊u(t)

julia> observed(sys)
6-element Vector{Equation}:
 int₊output₊u(t) ~ int₊x(t)
 int₊y(t) ~ int₊x(t)
 c2₊output₊u(t) ~ c2₊k
 c1₊output₊u(t) ~ c1₊offset + IfElse.ifelse(t < c1₊start_time, 0, c1₊amplitude*sin(c1₊phase + 6.283185307179586c1₊frequency*(t - c1₊start_time)))
 div₊output₊u(t) ~ div₊input1₊u(t) / div₊input2₊u(t)
 int₊u(t) ~ int₊input₊u(t)
```

It seems that during the structural transformations `isinput(var)` does make a huge difference and all variables with `input=true` are kept as state:
```julia
julia> states(sys)
4-element Vector{Term{Real, Base.ImmutableDict{DataType, Any}}}:
 int₊x(t)
 int₊input₊u(t)
 div₊input1₊u(t)
 div₊input2₊u(t)
```
What is the motivation behind that?

ValentinKaisermayer:
> It could be related. I recommend to not set flow variables to 0, since zero flow rate is often the singular case.

Unfortunately the tests never got that far.

ValentinKaisermayer:
see also https://github.com/SciML/ModelingToolkit.jl/issues/1577

YingboMa:
It's just because the system has a division by zero problem because of the initialization point.
```julia
julia> prob = ODEProblem(sys, Pair[int.x=>0.0, div.input2.u=>1.0], (0.0, 1.0))
ODEProblem with uType Vector{Float64} and tType Float64. In-place: true
timespan: (0.0, 1.0)
u0: 4-element Vector{Float64}:
 0.0
 0.0
 0.0
 1.0

julia> solve(prob, Rodas4())
```

ValentinKaisermayer:
Can this be inferred automatically? Since `u2` is connected to the constant with a value of 2 it should be 2. 
I guess in the first case, where the inputs do not end up in the state vector `alias_elimiate` takes care if it.

ValentinKaisermayer:
I will close this. `input=true` should not be required nor throw a warning - at least how it is implemented currently.

PR #56, opened 2022-05-21
Upgrade to new MTK

YingboMa:
DAE observed solution indexing is broken right now. https://github.com/SciML/ModelingToolkit.jl/pull/1575


ValentinKaisermayer:
The other option would be to to remove `der_T` from the thermal mass and keep the tests as is. The advantage would be that MTKSL Thermal would be in a usable state.

ChrisRackauckas:
Do that as a follow up?

PR #59, opened 2022-05-22
Add `smooth` sources in `Blocks` + Add `Square` and `Triangular` sources

ven-k:
- Adds `smooth` option for sources in Blocks. For `Sine`, `Cosine`, `ExpSine`, `Step`, `Ramp`

- Adds Square and Triangular sources

### DIFF
--- a/src/Blocks/sources.jl
+++ b/src/Blocks/sources.jl
@@ -1,3 +1,57 @@
+# Define and register smooth functions
+# These are "smooth" aka differentiable and avoid Gibbs effect
+# These follow: `offset` + `smooth_wave` * `smooth_step` with zero output for `t < start_time`
+function smooth_cos(x, δ, f, amplitude, ϕ, offset, start_time)
+    offset + amplitude * cos(2*π*f*(x - start_time) + ϕ) * smooth_step(x, δ, one(x), zero(x), start_time)
+end
+
+function smooth_damped_sin(x, δ, f, amplitude, damping, ϕ, offset, start_time)
+    offset + exp((start_time - x)*damping)*amplitude*sin(2*π*f*(x - start_time) + ϕ) * smooth_step(x, δ, one(x), zero(x), start_time)
+end
+
+function smooth_ramp(x, δ, height, duration, offset, start_time)
+    offset + height/(duration) * (smooth_xH(x, δ, start_time) - smooth_xH(x, δ, start_time+duration))
+end
+
+function smooth_sin(x, δ, f, amplitude, ϕ, offset, start_time)
+    offset + amplitude * sin(2*pi*f*(x - start_time) + ϕ) * smooth_step(x, δ, one(x), zero(x), start_time)
+end
+
+function smooth_square(x, δ, f, amplitude, offset, start_time)
+    offset + amplitude*2atan(sin(2π*(x - start_time)*f)/δ)/π * smooth_step(x, δ, one(x), zero(x), start_time)
+end
+
+function smooth_step(x, δ, height, offset, start_time)
+    offset + height*(atan((x - start_time)/δ)/π + 0.5)
+end
+
+function smooth_triangular(x, δ, f, amplitude, offset, start_time)
+    offset + amplitude * (1-2acos((1 - δ)sin(2π*(x - start_time)*f))/π) * smooth_step(x, δ, one(x), zero(x), start_time)
+end
+
+function smooth_xH(x, δ, tₒ)
+    0.5*(x-tₒ) * (1+((x-tₒ)/sqrt((x-tₒ)^2+δ^2)))
+end
+
+function square(x, f, amplitude, offset, start_time)
+    offset + (x > start_time) * (amplitude * (4*floor(f*(x - start_time)) - 2*floor(2*(x - start_time)*f) + 1))
+end
+
+function triangular(x, f, amplitude, offset, start_time)
+    p = 1/f # period
+    offset + (x > start_time) * (4 * amplitude * f * abs(abs((x - p/4 - start_time) % p) - p/2) - amplitude)
+end
+
+@register_symbolic smooth_cos(x, δ, f, amplitude, ϕ, offset, start_time)
+@register_symbolic smooth_damped_sin(x, δ, f, amplitude, damping, ϕ, offset, start_time)
+@register_symbolic smooth_ramp(x, δ, height, duration, offset, start_time)
+@register_symbolic smooth_sin(x, δ, f, amplitude, ϕ, offset, start_time)
+@register_symbolic smooth_square(x, δ, f, amplitude, offset, start_time)
+@register_symbolic smooth_step(x, δ, height, offset, start_time)
+@register_symbolic smooth_triangular(x, δ, f, amplitude, offset, start_time)
+@register_symbolic triangular(x, f, amplitude, offset, start_time)
+@register_symbolic square(x, f, amplitude, offset, start_time)
+
 """
 Generate constant signal.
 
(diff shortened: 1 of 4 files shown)


src/Blocks/sources.jl
```diff
@@ -34,12 +77,19 @@ function Sine(;name,
...
-    start_time=0)
+    start_time=0,
+    smooth=false)
 
     @named output = RealOutput()
     pars = @parameters offset=offset start_time=start_time amplitude=amplitude frequency=frequency phase=phase
+    equation = if smooth == false
+        offset + ifelse(t < start_time, 0, amplitude* sin(2*pi*frequency*(t - start_time) + phase))
+    else
+        δ = 1e-5
```
baggepinnen, line 88: Should this parameter be exposed and documented?

ValentinKaisermayer: I think it should

ven-k: Perhaps yes. Although having δ = 1e-6 or such makes a miniscule change to the result, it can upset high precision tests

src/Blocks/sources.jl
```diff
@@ -99,41 +159,87 @@ Generate ramp signal.
...
+
+    @named output = RealOutput()
+    pars = @parameters begin
+        offset=offset
+        amplitude=amplitude 
+        start_time=start_time 
+    end
+
+    eqs = [
+        output.u ~ smooth_square(t, δ, frequency, amplitude, offset, start_time)
```
ValentinKaisermayer, line 214: Is there no exact version?

src/Blocks/sources.jl
```diff
@@ -148,23 +254,64 @@ Generate exponentially damped sine signal.
...
+    pars = @parameters begin
+        offset=offset
+        amplitude=amplitude 
+        frequency=frequency 
+        start_time=start_time 
+    end
+
     eqs = [
-        output.u ~ offset + ifelse(t < start_time, 0, amplitude * exp(-damping * (t - start_time)) * sin(2*pi*frequency*(t - start_time) + phase))
+        output.u ~ smooth_triangular(t, δ, frequency, amplitude, offset, start_time)
```
ValentinKaisermayer, line 312: Is there no exact version?

ven-k:
If Triangular is going to be only smooth, I'm thinking, we should remove `start_time` and `offset` options

`start_time` shifts the wave but value isn't non-zero for t<st
When offset=0, start_time=1, amplitude=1, f=5<br>
A=1,f=5,st=1,o=0
<br>

Offset makes it very erratic 
When offset=1, start_time=1, amplitude=1, f=5 <br>
A=1,f=5,st=1,o=1

ven-k:
Square Wave
square-wave

src/Blocks/sources.jl
ven-k: When I had tried the solvers became unstable. Let me give another shot at it

ValentinKaisermayer: I would leave it in. It is an upstream problem than and not of the library. Also add tests but mark it as broken.

ValentinKaisermayer:
Is this an aliasing problem or is the implementation incorrect?

baggepinnen:
One option that can be provided alongside the smoothing option is to use callbacks to make sure the solver steps exactly at the discontinuities (in value or derivative etc.) of the signal. The symbolic callbacks should be able to handle these signals, but discrete callbacks may be more efficient given that we know the time points of the events beforehand.

ven-k:
For frequency=5, amplitude=1, offset = 2,  start_time = 1
<br>
Smooth triangular:
smooth-triangular
<br>
Triangular
triangular

----

For frequency=1, amplitude=2, offset=1, start_time=2.5
<br>
Smooth Square:
smooth-square
<br>
Square:
square

ValentinKaisermayer:
Cool!
The smooth square and exact square should have the same phase, for the same parameters, though.

ven-k:
> The smooth square and exact square should have the same phase

Yeah I had noted that. The last commit, I was working on, addresses exactly that

ven-k:
> One option that can be provided alongside the smoothing option is to use callbacks 

I will add this with a follow up PR

ven-k:
The thermal.jl tests are no longer broken. See the logs [here](https://github.com/SciML/ModelingToolkitStandardLibrary.jl/runs/6649725466?check_suite_focus=true#step:6:1198)
 (MTK@8.13.2 is used)

So can I go ahead & change `@test_broken` -> `@test` ?
cc @YingboMa

ValentinKaisermayer:
If you do, thx. But in a separate PR. 😄

ven-k:
Ok. 
Tests of this PR pass. And this is ready

ChrisRackauckas:
It looks like tests fail. Change test broken to test.

src/Blocks/sources.jl
```diff
@@ -145,26 +283,81 @@ Generate exponentially damped sine signal.
...
+             It uses a smoothing factor of `δ=1e-5`
+
+# Connectors:
+- `output`
+"""
+function Triangular(; name, amplitude=1.0, frequency=1.0,
+    offset=0.0, start_time=0.0, smooth=false)
+
+    if smooth
+        frequency > 25 && @warn "`frequency > 25` can lead to non-triangular wave pattern"
```
ValentinKaisermayer, line 339: What's up with this?

src/Blocks/sources.jl
```diff
@@ -1,3 +1,57 @@
...
+
+@register_symbolic smooth_cos(x, δ, f, amplitude, ϕ, offset, start_time)
+@register_symbolic smooth_damped_sin(x, δ, f, amplitude, damping, ϕ, offset, start_time)
+@register_symbolic smooth_ramp(x, δ, height, duration, offset, start_time)
+@register_symbolic smooth_sin(x, δ, f, amplitude, ϕ, offset, start_time)
+@register_symbolic smooth_square(x, δ, f, amplitude, offset, start_time)
+@register_symbolic smooth_step(x, δ, height, offset, start_time)
+@register_symbolic smooth_triangular(x, δ, f, amplitude, offset, start_time)
+@register_symbolic triangular(x, f, amplitude, offset, start_time)
+@register_symbolic square(x, f, amplitude, offset, start_time)
```
ChrisRackauckas, line 53: Why register these?

src/Blocks/sources.jl
baggepinnen: Perhaps the smoothing constants need to be frequency dependent?

ven-k: f=25, 26, 30

For f>25, the waveform is totally erratic

>Perhaps the smoothing constants need to be frequency dependent?

I tried these:
```
amplitude * (1-2acos((1 - f* δ)sin(2π*(x - start_time)*f))/π)
amplitude * (1-2acos((1 - δ/f)sin(2π*(x - start_time)*f))/π)
amplitude * (1-2acos((1 - δ)sin(2π*(x - start_time)*f*δ))/π)
amplitude * (1-2acos((1 - δ)sin(2π*(x - start_time)*f/δ))/π)
amplitude * (1-2acos(sin(2π*(x - start_time)*f)/π)
```

None of these relations gave a better result.
I think it comes down to trigonometric functions used. Is there any source I can look at for this?

src/Blocks/sources.jl
ven-k: Earlier we didn't assign parameters while declaring them inside components. So `@register` was necessary.

But yeah, now its no longer required. Removed

src/Blocks/sources.jl
ValentinKaisermayer: Can you try evaluating the functions explicitly? Maybe it is just a solver issue.

ChrisRackauckas: This is still open which is why I didn't merge. What's the consensus here? I didn't look into it.

ven-k: Aha, the real issue is the stepsize. And this is present in both `smooth=true/false` wave.

for `t = 0:0.001:1`

tri-smooth-f=26


Should I mention a note on this in docstring?

(Will remove the `@warn`)

ChrisRackauckas: Adaptivity should handle it?

ven-k: Somehow it didn't. For whenever `offset!=0` similar issue crops up. `saveat=0.01` saves. ([Here](https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/55668b14c365562b592be1fb8d805e4edc860f16/test/Blocks/sources.jl#L270))

PR #66, opened 2022-06-07
Add `Short` in Analog and `TwoPort` in Electrical

ven-k:
And, use `TwoPort` in `IdealOpAmp`


ValentinKaisermayer (approved):
Export `Short`

ven-k:
It's already exported [here](https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/main/src/Electrical/Electrical.jl#L33); it wasn't defined

PR #75, opened 2022-06-10
adds missing rotational components

ValentinKaisermayer:
# Added Components:
## Components
- `RotationalFriction`
## Sources
- `Torque`
- `Speed`
## Sensors
- `AngleSensor`
- `SpeedSensor`
- `TorqueSensor`
- `RelSpeedSensor`


ValentinKaisermayer:
Closes #28

test/Mechanical/rotational.jl
```diff
@@ -107,9 +107,11 @@ end
 
     @test_throws Any prob = ODAEProblem(sys, Pair[], (0, 1.0))
 
-    prob = DAEProblem(sys, D.(states(sys)) .=> 0.0, Pair[], (0, 10.0))
-    sol = solve(prob, DFBDF())
-    @test sol.retcode == :Success
+    @test_skip begin
+        prob = DAEProblem(sys, D.(states(sys)) .=> 0.0, Pair[], (0, 10.0)) # ArgumentError: SymbolicUtils.Term{Real, Base.ImmutableDict{DataType, Any}}[idealGear₊phi_aˍt(t)] are missing from the variable map.
```
ValentinKaisermayer, line 111: @YingboMa I think, if MTK adds variables it should handle the initial conditions on its own. 
Does this correspond to `D(idealGear.phi_a)`?

YingboMa: Yes, `xˍt` mean `D(x)`. MTK cannot make up initial conditions though. Maybe we need an interface to programmatically tell the user added variables instead of communicating this in an error.

ValentinKaisermayer: If I supply `D(idealGear.phi_a)` would it work?

YingboMa: Yes

ValentinKaisermayer: The error should simply state that. At least the user knows what to do then.

ValentinKaisermayer: I find this very unintuitiv. Now a user has to supply an initial value for the derivative of a variable but via `u0`:
```julia
prob = DAEProblem(sys, (D.(states(sys)) .=> 0.0, [D(idealGear.phi_a) => 0.0], (0, 10.0))
```

Also the system is unbalanced after structual_simplify: 
```julia
julia> sys
Model model with 5 equations
States (6):
  inertia3₊phi(t) [defaults to 0.0]
  inertia3₊w(t) [defaults to 0.0]
⋮
Parameters (13):
```

baggepinnen: I think this is something new that has started happening, MTK seems to require initial values for higher-order derivatives lately than it used to require
https://github.com/SciML/ModelingToolkit.jl/issues/1628

baggepinnen: Just spoke to Yingbo, a fix might be coming up real soon

ChrisRackauckas:
Conflicts

ValentinKaisermayer:
I have run `format(".")`, no idea why it still complains.

YingboMa (approved):

ValentinKaisermayer:
The PR does not even touch the file it complains about.

ValentinKaisermayer:
> Are you sure you are using JuliaFormater 1.0?

v1.0.3

ValentinKaisermayer:
Tried with v1.0.5
```julia
julia> format(".")
true
```

baggepinnen:
@ValentinKaisermayer check that the Julia environment from which you call `JuliaFormatter` is using you `dev`ed version of MTKStdlib, I just ran into a similar problem where JuliaFormatter ran on the latest released version of MTK instead of my local copy :P

ValentinKaisermayer:
I'm pretty sure it is the correct folder. The package folder is read only, isn't it?

YingboMa (approved):

PR #78, opened 2022-06-15
Adds hydraulic library

ValentinKaisermayer:
There seems to be still an issue, maybe with  the solver, I'm not sure.
The mass flow rate should get to zero when the pressure of the volume is equal to the pressure of the source.

## Rodas4
hydraulic

## Tsit5
hydraulic

Since many components need some sort of media properties, I used the variable scope to lift them into the overall model where they can be centrally defined. Maybe there is a better way to do this. Modelica and Simscape do this via a special block that is present in the model.

### DIFF
--- a/src/Hydraulic/Hydraulic.jl
+++ b/src/Hydraulic/Hydraulic.jl
@@ -1,0 +1,16 @@
+module Hydraulic
+using ModelingToolkit, OrdinaryDiffEq
+
+@parameters t
+D = Differential(t)
+
+export HydraulicPort, FluidProperties
+include("utils.jl")
+
+export LocalRestriction, ConstantVolume, Reservoir
+include("components.jl")
+
+export PressureSource, MassFlowRateSource
+include("sources.jl")
+
+end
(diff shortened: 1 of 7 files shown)


ValentinKaisermayer:
Works fine with smaller solver tolerance:
hydraulic

ChrisRackauckas:
Conflicts.

Yes. looks like a standard numerical instability due to large dts.

dodoplus:
@ValentinKaisermayer just curious: why are you modeling changes in density due to pressure but ignore temperature entirely? E.g. for water, 10K change in temperature is more significant than 10MPa change in pressure.

ValentinKaisermayer:
Because this library is supposed to model fluid systems where the change in temperature is negligible.  This allows to avoid stream connectors all together. However, if this is an insufficient assumption for the modeling needs at hand, a `ThermoFluid` domain is needed, i.e. #87.

[WSM](https://reference.wolfram.com/system-modeler/libraries/Hydraulic/Hydraulic.html) and [Simscape](https://www.mathworks.com/help/physmod/simscape/isothermal-liquid-models.html)  have such a library.

ValentinKaisermayer:
I guess this is obsolete with #159

PR #83, opened 2022-06-21
add optional operating point in StateSpace

baggepinnen:
Useful if the `StateSpace` system comes from a linearized nonlinear model.


src/Blocks/continuous.jl
```diff
@@ -417,16 +417,28 @@ function LimPID(; name, k=1, Ti=false, Td=false, wp=1, wd=1,
...
+```math
+ẋ = Ax + B(u - u0)
+y = Cx + D(u - u0) + y0
+```
+For a nonlinear system
+```math
+ẋ = f(x, u)
+y = h(x, u)
+```
+linearized around the operating point `x₀, u₀`, we have `y0, u0 = h(x₀, u₀), u₀`.
```
ValentinKaisermayer, line 439: Writing two equations as tuple is confusing I think.

src/Blocks/continuous.jl
```diff
@@ -417,16 +417,28 @@ function LimPID(; name, k=1, Ti=false, Td=false, wp=1, wd=1,
...
 ẋ = Ax + Bu
 y = Cx + Du
 ```
 Transfer functions can also be simulated by converting them to a StateSpace form.
+
+`y0` and `u0` can be used to set an operating point, providing them changes the dynamics to
+```math
+ẋ = Ax + B(u - u0)
+y = Cx + D(u - u0) + y0
+```
```
ValentinKaisermayer, line 433: This is actually no longer an LTI system but affine.

src/Blocks/continuous.jl
```diff
@@ -417,16 +417,28 @@ function LimPID(; name, k=1, Ti=false, Td=false, wp=1, wd=1,
...
 
 A linear, time-invariant state-space system on the form.
-```
+```math
 ẋ = Ax + Bu
 y = Cx + Du
 ```
 Transfer functions can also be simulated by converting them to a StateSpace form.
+
+`y0` and `u0` can be used to set an operating point, providing them changes the dynamics to
```
baggepinnen, line 429:
```suggestion
`y0` and `u0` can be used to set an operating point, providing them changes the dynamics from an LTI system to the affine system 
```

baggepinnen: @ValentinKaisermayer is this a good clarification?

ValentinKaisermayer: Looks good!

YingboMa (commented):
Why `x_start` but `u0` and `y0`?

baggepinnen:
> Why x_start but u0 and y0?

`x_start` denotes a start value, whereas `x0, u0` denote nominal/operating points (a rather commonly used notation to describe the operating point for delta-coordinates $\Delta x = x - x_0$).

PR #85, opened 2022-06-21
Add translational library

kianmolani:
This PR is linked to issue #84


.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
YingboMa, line 1: You can add `.vscode` and `.DS_Store` in global git ignore setting. Check out https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

test/Mechanical/translational.jl
```diff
@@ -0,0 +1,36 @@
+import ModelingToolkitStandardLibrary.Blocks
+include("/Users/kianmolani/Documents/Project Files/Codebases/ModelingToolkitStandardLibrary.jl/src/Mechanical/Translational/Translational.jl")
+using ModelingToolkit, OrdinaryDiffEq, Test
+using Plots
```
YingboMa, line 4: Don't import Plots in test. CI cannot actually see plots.

src/Mechanical/Translational/Translational.jl
```diff
@@ -0,0 +1,19 @@
...
+using OffsetArrays
+using ...Blocks: RealInput, RealOutput
+
+@parameters t
+D = Differential(t)
+
+export Flange
+include("utils.jl")
+
+export Fixed, Inertia, Spring, Damper, IdealGear
```
baggepinnen, line 13:
```suggestion
export Fixed, Mass, Spring, Damper, IdealGear
```

src/Mechanical/Translational/Translational.jl
```diff
@@ -0,0 +1,19 @@
+module Translational
+
+using ModelingToolkit, Symbolics, IfElse, OrdinaryDiffEq
+using OffsetArrays
```
ValentinKaisermayer, line 4: Remove not needed packages

ValentinKaisermayer:
Doc strings are missing and actual tests

baggepinnen:
Have a look at https://github.com/baggepinnen/AutomaticDocstrings.jl to quickly generate docstring stubs, helps a lot when you have a lot of docstrings to generate

test/Mechanical/translational.jl
```diff
@@ -0,0 +1,36 @@
+import ModelingToolkitStandardLibrary.Blocks
+include("/Users/kianmolani/Documents/Project Files/Codebases/ModelingToolkitStandardLibrary.jl/src/Mechanical/Translational/Translational.jl")
```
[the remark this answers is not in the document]
kianmolani: How can I add the Translational module to the package and be able to import it using `use ModelingToolkitStandardLibrary.Mechanical.Translational`? Do I have to first merge this PR? I have tried some possible solutions but without success.

kianmolani (commented):
I will delete the output folder before the PR is merged - I've just included the figures here for your reference (though lmk if there's a better way to share output results and plots). Also, I don't know whether we want to run tests against Modelica simulation data. I certainly think it's useful as an initial sanity check, but don't know if it's appropriate to include as part of continuous testing.

src/Mechanical/Translational/utils.jl
```diff
@@ -0,0 +1,154 @@
...
+    sts = @variables begin
+        s(t)
+        f(t), [connect=Flow]
+    end
+    ODESystem(Equation[], t, sts, [], name=name, defaults=Dict(s=>0.0, f=>0.0))
+end
+Base.@doc """
+    Flange(;name)
+
+1-dim. translational flange of a shaft.
```
ValentinKaisermayer, line 11:
```suggestion
1-dim. translational flange.
```

src/Mechanical/Translational/utils.jl
```diff
@@ -0,0 +1,154 @@
...
+
+# States:
+- `s`: [m] Absolute position of flange
+- `f`: [N] Cut force in the flange
```
ValentinKaisermayer, line 15:
```suggestion
- `f`: [N] Cut force into the flange
```

src/Mechanical/Translational/utils.jl
```diff
@@ -0,0 +1,154 @@
...
+    sts = @variables begin
+        s(t)
+        f(t), [connect=Flow]
+    end
+    ODESystem(Equation[], t, sts, [], name=name, defaults=Dict(s=>0.0, f=>0.0))
+end
+Base.@doc """
+    Support(;name)
+
+Support/housing of a 1-dim. translational shaft
```
ValentinKaisermayer, line 28:
```suggestion
Support/housing 1-dim. translational flange.
```

src/Mechanical/Translational/utils.jl
```diff
@@ -0,0 +1,154 @@
...
+
+# States:
+- `s`: [m] Absolute position of the support/housing
+- `f`: [N] Reaction force in the support/housing
```
ValentinKaisermayer, line 32:
```suggestion
- `f`: [N] Cut force into the flange
```

src/Mechanical/Translational/utils.jl
```diff
@@ -0,0 +1,154 @@
...
+""" Support
+
+"""
+    PartialCompliant(;name, s_rel_start=0.0, f_start=0.0)
+
+Partial model for the compliant connection of two translational 1-dim. shaft flanges.
```
ValentinKaisermayer, line 38:
```suggestion
Partial model for the compliant connection of two translational 1-dim. flanges.
```

src/Mechanical/Translational/utils.jl
```diff
@@ -0,0 +1,154 @@
...
+
+# Parameters:
+- `s_rel_start`: [m] Initial relative distance
```
ValentinKaisermayer, line 41:
```suggestion
- `s_rel_start`: [m] Initial relative distance between the flanges 
```

src/Mechanical/Translational/utils.jl
```diff
@@ -0,0 +1,154 @@
...
+        flange_b.f ~ f
+        flange_a.f ~ -f
+    ]
+    return compose(ODESystem(eqs, t, sts, []; name=name), flange_a, flange_b)
+end
+
+"""
+    PartialCompliantWithRelativeStates(;name, s_rel_start=0.0, v_rel_start=0.0, a_rel_start=0.0, f_start=0.0)
+
+Partial model for the compliant connection of two translational 1-dim. shaft flanges where the relative position and speed are used as preferred states
```
ValentinKaisermayer, line 66:
```suggestion
Partial model for the compliant connection of two translational 1-dim. flanges.
```

test/Mechanical/translational.jl
```diff
@@ -0,0 +1,58 @@
+import ModelingToolkitStandardLibrary.Blocks
+include(pwd() * "/src/Mechanical/Translational/Translational.jl")
```
ValentinKaisermayer, line 2:
```suggestion
using Modelingtoolkitstandardlibrary.Mechanical.Translational
```

test/Mechanical/translational.jl
```diff
@@ -0,0 +1,58 @@
...
+    mass_v_julia = sol[mass.v]
+    mass_a_julia = sol[mass.a]
+
+    # Define and run tests
+
+    @test sol.retcode == :Success
+    @test mass_v_julia[end] ≈ 0 atol = 1e-3 # all energy has dissipated
+    @test mean(broadcast(abs, (mass_s_julia - mass_s_modelica))) ≈ 0 atol = 1e-5 # jl and modelica s vs. t results are approx equal
+    @test mean(broadcast(abs, (mass_v_julia - mass_v_modelica))) ≈ 0 atol = 1e-4 # jl and modelica v vs. t results are approx equal
+    @test mean(broadcast(abs, (mass_a_julia - mass_a_modelica))) ≈ 0 atol = 1e-2 # jl and modelica a vs. t results are approx equal
```
ValentinKaisermayer, line 56: Can you make the test self-sufficient?

src/Mechanical/Translational/utils.jl
```diff
@@ -0,0 +1,154 @@
...
+    else
+        eqs = [s_support ~ 0]
+    end
+    return compose(ODESystem(eqs, t, [s_support], []; name=name), sys)
+end
+
+"""
+    PartialElementaryTwoFlangesAndSupport2(;name, use_support=false)
+
+Partial model for a component with two translational 1-dim. shaft flanges and a support used for textual modeling, i.e., for elementary models
```
ValentinKaisermayer, line 130:
```suggestion
Partial model for a component with two translational 1-dim. flanges and a support used for textual modeling, i.e., for elementary models
```

src/Mechanical/Translational/Translational.jl
```diff
@@ -0,0 +1,21 @@
+"""
+Library to model 1-dimensional, translational mechanical systems
```
ValentinKaisermayer, line 2:
```suggestion
Library to model 1-dimensional, translational mechanical components.
```

ValentinKaisermayer (commented):
Can you add the `connectors` in the doc strings as well? And name the states `s(t)` to be consistent with the rest of the library.

If you want to add images, simply drag them into a comment or an issue.

kianmolani:
> Can you add the `connectors` in the doc strings as well? And name the states `s(t)` to be consistent with the rest of the library.
> 
> If you want to add images, simply drag them into a comment or an issue.

Thank you for the suggestions, Valentin. The connectors in `utils.js` already have doc strings added (I mirrored the documentation from the same file in the Rotational lib). Are you referring to something else? Also, what do you mean about naming the states `s(t)`? The position state is already denoted by `s`.

ValentinKaisermayer:
I meant a `# Connectors:` section in the doc strings, listing the connectors of the components. You can have a look at the doc strings in the other modules.

test/Mechanical/translational.jl
kianmolani: As written, this line returns the error `Translational not defined`. Is this because the Translational module has not been added to the package using the proper procedures (that is, using JL's package manager)? Note that when creating the Translational library, I merely added a new folder to the repo and proceeded to add the requisite lines of code.

test/Mechanical/translational.jl
kianmolani: I can't think of a way to do this without importing the Modelica simulation results. Please let me know if you see a better way. And also, what are your thoughts in general about running these sort of tests? Do you think they better serve as one-offs for sanity checking (i.e. not to be repeated again), or for continuous testing? Please see my original comments on this issue: https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/85#pullrequestreview-1018686760

kianmolani:
There appear to be issues with our simulated results for a mass-spring-damper model. Julia and Modelica simulation results do not match, and the difference is very apparent. I've attached below screenshots of Julia and Modelica simulation results of our mass position vs. time for underdamped, overdamped, and critically damped systems. For the underdamped system, the difference appears to be accounted for with an offset of about 0.5, as well as a slight phase shift. For the overdamped and critically damped systems, the solutions possess different equilibrium states.

Screen Shot 2022-07-14 at 11 56 20 AM

Screen Shot 2022-07-14 at 11 58 02 AM

Screen Shot 2022-07-14 at 11 59 15 AM

I've ensured that both Modelica and Julia simulations are initialized in the same way. Note that simulation results between Modelica and Julia for a mass-damper model are very close (tested across a range of parameter values).

This last screenshot shows how I've defined my system connections for a mass-spring-damper model:

Screen Shot 2022-07-14 at 9 46 16 AM

docs/src/tutorials/mass_spring_damper.md
```diff
@@ -0,0 +1,125 @@
...
+s_start = 3
+v_start = 10
+m = 1
+c = 10
+d = 1
+
+@named fixed = Fixed(s0=s0)
+@named mass = Mass(m=m, s_start=s_start, v_start=v_start)
+@named damper = Damper(d=d)
+@named spring = Spring(c=c, s_rel0=s_rel0)
```
baggepinnen, line 28:
```suggestion
@named fixed = Fixed(; s0)
@named mass = Mass(; m, s_start, v_start)
@named damper = Damper(; d)
@named spring = Spring(; c, s_rel0)
```
In case you didn't know, if your variable shares the same name as a keyword argument (a common case), you do not need to repeat the name. You do need to have the semi-colon first though to indicate that keyword arguments are coming.

baggepinnen:
Could the difference between julia and modelica lie in the initialization? Is there a difference already at the first point saved in the solution?

baggepinnen:
```julia
s0 = 4.5
s_rel0 = 1
s_start = 3

**Comment by baggepinnen** [comment]
Let me see if I understand this correctly, the fixed point is at `s = 4.5` and the spring has resting length `s_rel0 = 1`, then I'd expect the steady-state position to be either 5.5 or 3.5 depending on the direction the spring is pointing in. It thus looks like the Julia solution is correct since it comes to rest at 3.5? I assume there's no gravity or other external forces acting anywhere? Is the length of the spring and all other params the same in modelica?

**Comment by kianmolani** [comment]
> Let me see if I understand this correctly, the fixed point is at `s = 4.5` and the spring has resting length `s_rel0 = 1`, then I'd expect the steady-state position to be either 5.5 or 3.5 depending on the direction the spring is pointing in. It thus looks like the Julia solution is correct since it comes to rest at 3.5? I assume there's no gravity or other external forces acting anywhere? Is the length of the spring and all other params the same in modelica?

No other forces and all other params confirmed to be the same.

The issue might be that Modelica takes in an additional parameter that we do not: mass length (set to 1 in this simulation). Because Modelica also has the parameter `s_start` (the initial value of the absolute position of the mass), and because this parameter is set to 3, it means that Modelica is tracking the center of this mass. Therefore, the equilibrium state at `x = 3` makes sense for the Modelica simulation.

However, _we_ do not take in a parameter for mass length, and so I guess that we're treating the mass as a point mass? In any case, initializing `s_start` to also equal 3 in our simulation (given that we do not have a mass length) might be problematic in that there's a discontinuity in the connections. I've drawn two schematics to illustrate this point:

For Modelica:

image

For Julia:

image

**Comment by kianmolani** [comment]
Indeed, I've just confirmed that setting the spring length to 1.5m in our Julia simulation in order to account for this "gap" yields matching simulation results.

**Comment by kianmolani** [comment]
Perhaps mass length is a property that we should include as well.

**Comment by kianmolani** [comment]
> You should remove `.vscode` and `.DS_Store`, then the PR is good to go after formatting.

All `.vscode` and `.DS_Store` files are now deleted

**Comment by YingboMa** [comment]
Format failure.

File: docs/src/tutorials/mass_spring_damper.md [hunk 18]
```diff
@@ -0,0 +1,125 @@
...
+
+This tutorial demonstrates the response of a mass-spring-damper system. Rather than applying an external force, the mass is initialized with some positional offset from equilibrium and some non-zero velocity (system responses under such conditions are commonly referred to as "unforced" responses). Assuming a lossless system, a mass-spring system in the absence of a damper will continually oscillate at some natural frequency. The addition of a damper will result in (1) underdamped, (2) overdamped, or (3) critically damped motion, depending on the relative values of our mass (`m`), spring constant (`c`), and damping constant (`d`). In this first example, we will demonstrate an underdamped system. Values for `m`, `c`, and `d` were chosen such that the roots of the polynomial `s^2 + d/m * s + c/m` are complex conjugates (the equivalent condition for underdamped motion is `d < sqrt(4 * m * c)`). In the case of underdamped systems, our mass will undergo damped oscillations. This damping corresponds to system energy loss, and the amplitudes of said oscillations will decrease towards zero as time approaches infinity.
+
+Let's start by importing what we need.
+
+```@example
```
ValentinKaisermayer, line 7:
```suggestion
```@example translational 
```

docs/src/tutorials/mass_spring_damper.md
```diff
@@ -0,0 +1,125 @@
...
+using ModelingToolkitStandardLibrary.Mechanical.Translational, ModelingToolkit, OrdinaryDiffEq, Plots
+import ModelingToolkitStandardLibrary.Blocks
+```
+
+Now let's initialize our system state and define our components and connections. Here, `s0` is the fixed offset position of our flange housing, `s_rel0` is our unstretched spring length, and `s_start` and `v_start` are the initial values of the absolute position and linear velocity of our sliding mass, respectively. Finally, `m`, `c`, and `d` are our mass, spring, and damping constants, respectively. All units are in SI units.
+
+```@parameters t
```
ValentinKaisermayer, line 14:
```suggestion
```@example translational 
@parameters t
```

docs/src/tutorials/mass_spring_damper.md
```diff
@@ -0,0 +1,125 @@
...
+
+connections = [
+    connect(mass.flange_b, damper.flange_a, spring.flange_a)
+    connect(damper.flange_b, spring.flange_b, fixed.flange)
+]
+```
+
+Before solving our ODE system, let us ensure that the condition for underdamped motion is met.
+
+```
```
ValentinKaisermayer, line 38:
```suggestion
```@example translational 
```

docs/src/tutorials/mass_spring_damper.md
```diff
@@ -0,0 +1,125 @@
...
+@assert(d < sqrt(4 * m * c))
+```
+
+Now let's solve our ODE system and plot our results (note that we are plotting the position of our mass as a function of time).
+
+```@named model = ODESystem(connections, t, systems=[fixed, mass, damper, spring])
```
ValentinKaisermayer, line 44:
```suggestion
```@example translational 
@named model = ODESystem(connections, t, systems=[fixed, mass, damper, spring])
```

docs/src/tutorials/mass_spring_damper.md
```diff
@@ -0,0 +1,125 @@
...
+    ylabel = "Position (m)")
+
+savefig("mass_spring_damper_underdamped.png"); nothing # hide
+```
+
+!["Position vs. Time Plot of a Mass in an Underdamped Mass-Spring-Damper System](mass_spring_damper_underdamped.png)
+
+We'll now demonstrate a mass-spring-damper system that is overdamped. For such systems, `d > sqrt(4 * m * c)`, and the zeros of our characteristic polynomial are real. Let us re-initialize our `m`, `c`, and `d` constants accordingly, and re-define our components and connections.
+
+```
```
ValentinKaisermayer, line 61:
```suggestion
```@example translational 
```

docs/src/tutorials/mass_spring_damper.md
```diff
@@ -0,0 +1,125 @@
...
+
+connections = [
+    connect(mass.flange_b, damper.flange_a, spring.flange_a)
+    connect(damper.flange_b, spring.flange_b, fixed.flange)
+]
+```
+
+Let us ensure that the condition for overdamped motion is met.
+
+```
```
ValentinKaisermayer, line 76:
```suggestion
```@example translational 
```

docs/src/tutorials/mass_spring_damper.md
```diff
@@ -0,0 +1,125 @@
...
+@assert(d > sqrt(4 * m * c))
+```
+
+Let's solve our ODE system and plot our results. Observe that for overdamped systems, motion is non-oscillatory.
+
+```@named model = ODESystem(connections, t, systems=[fixed, mass, damper, spring])
```
ValentinKaisermayer, line 82:
```suggestion
```@example translational 
@named model = ODESystem(connections, t, systems=[fixed, mass, damper, spring])
```

docs/src/tutorials/mass_spring_damper.md
```diff
@@ -0,0 +1,125 @@
...
+    ylabel = "Position (m)")
+
+savefig("mass_spring_damper_overdamped.png"); nothing # hide
+```
+
+!["Position vs. Time Plot of a Mass in an Overdamped Mass-Spring-Damper System](mass_spring_damper_overdamped.png)
+
+Our last demonstration will be of a mass-spring-damper system that is critically damped. The condition for critically damped motion is `d = sqrt(4 * m * c)`, where the roots of our characteristic polynomial are both equal to `d/2m`. Again, let us re-initialize and re-define our system state, components, and connections, solve our ODE system, and plot our results. Observe that for critically damped systems, motion is non-oscillatory.
+
+```
```
ValentinKaisermayer, line 99:
```suggestion
```@example translational 
```

ValentinKaisermayer (changes requested):
Other than the docs it looks good. 👍

bradcarman:
Hi, I'm reviewing the code and I noticed that the Flange connector uses absolute position for the state variable.  Might I suggest changing this to velocity?  The reason for this is so we have less matching parameters to set.  Let me show an example...

The below code requires that each of the 3 components have a position parameter set, and `s_rel0` is really just `s0-s_start`, so it can already be seen as redundant.

```julia
s0 = 4.5 # fixed offset position of flange housing
s_start = 3 # initial value of absolute position of sliding mass
s_rel0 = 1.5 # unstretched spring length
v_start = 10 # initial value of absolute linear velocity of sliding mass

@named fixed = Fixed(; s0)
@named mass = Mass(; m=1, s_start, v_start)
@named spring = Spring(; c=10, s_rel0)

eqs = [
    connect(mass.flange_a, spring.flange_a)
    connect(spring.flange_b, fixed.flange)
]

@named model = ODESystem(eqs, t, [], []; systems=[fixed, mass, spring])

prob = ODEProblem(sys, [], (0, 10.0))
sol = solve(prob, ImplicitMidpoint(), dt=0.01)
```
Plotting sol[mass.s]...
image

If I remove the position parameters from `Fixed` and `Spring`, you can see we get the incorrect result...

```julia
@named fixed = Fixed()
@named mass = Mass(; m, s_start, v_start)
@named spring = Spring(; c)
```
image

Now, for comparison (using #107) we can see that only the `Body` position is needed, which to me makes more sense because we really only need to track absolute positions of actual moving rigid bodies.  Additionally the `Fixed` component is defined by simply setting the port velocity to 0.

```julia
x0 = 3 # initial value of absolute position of sliding mass
v0 = 10 # initial value of absolute linear velocity of sliding mass

function Fixed(; name)
    @named port = Port()
    eqs = [port.v ~ 0]
    return compose(ODESystem(eqs, t, [], []; name = name), port)
end

@named fixed = Fixed()
@named mass = Body(; m=1, x0, v0)
@named spring = Spring(; k=10)

eqs = [
    connect(mass.port, spring.port_a)
    connect(spring.port_b, fixed.port)
]

@named model = ODESystem(eqs, t, [], []; systems=[fixed, mass, spring])

prob = ODEProblem(sys, [], (0, 10.0))
sol = solve(prob, ImplicitMidpoint(), dt=0.01)
```

image

This is also how other acausal modeling tools I know of define their thru and across (flow) variables for translational components.

baggepinnen:
> Might I suggest changing this to velocity?

The intention is to follow modelica stdlib, if we're deviating from them, it's probably an oversight and should be fixed :)

ValentinKaisermayer:
Although, Simscape uses velocity and angular velocity for its mechanical domains. 

In the example you show I think you do not need to specify `s_rel0` for the spring. It should be fully determined by the mass and fixed position.

bradcarman:
In this example the spring starts at it's natural length (i.e. uncompressed), therefore one should not need to specify `s_rel0`.  By using position rather than velocity, this overcomplicates the model and now one must specify both the position of the `Fixed` component as well as `s_rel0` for the spring.  

Additionally, because the `Flange` is defined with `s` defaulting to 0, we now introduce a requirement that the model must solve for it's initial states.  For a small model, this is not such a big deal, but for the large models I'm dealing with, initialization is something I'd like to avoid for 1) I need my model to start immediately, waiting for initialization is not an option for real time control, 2) I'd like to guarantee my states.

If we run this simple model without initialization and simplification, it can be seen that the result is not good.  

image

Full code below...

```julia
using ModelingToolkitStandardLibrary.Mechanical.Translational, ModelingToolkit,
      OrdinaryDiffEq, Test
using Setfield

@parameters t
D = Differential(t)

s0 = 4.5 # fixed offset position of flange housing
s_start = 3 # initial value of absolute position of sliding mass
s_rel0 = 1.5 # unstretched spring length
v_start = 10 # initial value of absolute linear velocity of sliding mass

@named fixed = Fixed(; s0)
@named mass = Mass(; m=1, s_start, v_start)
@named spring = Spring(; c=10, s_rel0)

eqs = [
    connect(mass.flange_a, spring.flange_a)
    connect(spring.flange_b, fixed.flange)
]

@named model = ODESystem(eqs, t, [], []; systems=[fixed, mass, spring])

sys = expand_connections(model)
@set! sys.eqs = [(typeof(eq.lhs) != ModelingToolkit.Connection) & !ModelingToolkit._iszero(eq.lhs) & !ModelingToolkit.isdifferential(eq.lhs) ? 0 ~ eq.rhs - eq.lhs : eq for eq in sys.eqs]

prob = ODEProblem(sys, [], (0, 2.0))
sol = solve(prob, ImplicitMidpoint(), dt=0.01, initializealg=NoInit())

using CairoMakie
begin
    f,a,=lines(sol.t, sol[mass.s], label="sol[mass.s]");    
    lines!(a, sol.t, sol[mass.flange_a.s], label="sol[mass.flange_a.s]")
    lines!(a, sol.t, sol[spring.flange_a.s], label="sol[spring.flange_a.s]", linestyle=:dash)
    lines!(a, sol.t, sol[spring.flange_b.s], label="sol[spring.flange_b.s]")
    lines!(a, sol.t, sol[fixed.flange.s], label="sol[fixed.flange.s]", linestyle=:dash)
    a.xlabel="time [s]"; a.ylabel="position [m]"; 
    Legend(f[1,2], a)
    f
end
```

Additionally we can see that `spring.s_rel` is not automatically resolved

image


In contrast, by using velocity and force for the translational Connection, we only need to specify the position of the Body/Mass, and no initialization is required (using #107)...

```julia
using ModelingToolkitStandardLibrary.Mechanical.Translational, ModelingToolkit,
      OrdinaryDiffEq, Test
using Setfield

@parameters t
D = Differential(t)

x0 = 3 # initial value of absolute position of sliding mass
v0 = 10 # initial value of absolute linear velocity of sliding mass

function Fixed(; name)
    @named port = Port()
    eqs = [port.v ~ 0]
    return compose(ODESystem(eqs, t, [], []; name = name), port)
end

@named fixed = Fixed()
@named mass = Body(; m=1, x0, v0)
@named spring = Spring(; k=10)

eqs = [
    connect(mass.port, spring.port_a)
    connect(spring.port_b, fixed.port)
]

@named model = ODESystem(eqs, t, [], []; systems=[fixed, mass, spring])

sys = expand_connections(model)
@set! sys.eqs = [(typeof(eq.lhs) != ModelingToolkit.Connection) & !ModelingToolkit._iszero(eq.lhs) & !ModelingToolkit.isdifferential(eq.lhs) ? 0 ~ eq.rhs - eq.lhs : eq for eq in sys.eqs]
prob = ODEProblem(sys, [], (0, 2.0))
sol = solve(prob, ImplicitMidpoint(), dt=0.01, initializealg=NoInit())

using CairoMakie
begin
    f,a,=lines(sol.t, sol[mass.x], label="sol[mass.x]");    
    a.xlabel="time [s]"; a.ylabel="position [m]"; 
    Legend(f[1,2], a)
    f
end
```

image

bradcarman:
Making my point in another way, one should be able to define the `Fixed` component such that position is not needed, we simply specify that the change in position is zero...

```julia
function Fixed(; name)
    @named flange = Flange()
    eqs = [D(flange.s) ~ 0]
    return compose(ODESystem(eqs, t, [], []; name = name), flange)
end
``` 

But, unfortunately this will not work in the end because `connect( fixed.flange, component.flange )` will give the equation...

```julia
fixed.flange.s ~ component.flange.s
```

Therefore, we are forced to specify the position for every single `Flange` in the model.  But in theory, it should not be necessary to specify the position for a `Fixed` component, this is really overly defining the model.

Firionus:
@bradcarman How would you express in your approach a spring that has non-zero length in its relaxed state? 
That would certainly be a important for someone that wants to integrate with visualization or if someone wants to model something precisely the way they imagine their coordinate system.

Firionus:
> And to answer @Firionus, for visualization, if we are plotting the position states, the only states tracked by the model are actual moving bodies/masses. So if we have 2 bodies connected by a spring, then we can plot both body positions.

This doesn't hold true if the spring is connected to a fixed element. Also, visualizing springs with a finite relaxed length is much closer to a real mechanical setup than springs that are infinitely small sometimes, don't you think? 

> One final point (then I'll shut up, ha ha), the necessity that one must specify a spring length for the Spring component is silly when one is considering models that really just deal with stiffness.

If you find it silly you can just leave it at the default of `s_rel0 = 0.0` and achieve that. I realize you'll still have to define the position of fixed points but I think this is a small price to pay for gaining better modelling of real life with the absolute position approach. 

Ultimately, it comes down to the choice of programming interface. Should a "mechanical component" only keep track of velocities and forces, i.e. its energy exchange, or should it also track its absolute position. The Modelica Standard Library (MSL 3.2.3) goes with the second approach:

```Modelica
connector Flange "One-dimensional translational flange"

  SI.Position s "Absolute position of flange";
  flow SI.Force f "Cut force directed into flange";
  annotation (...);
end Flange;
```
And since @baggepinnen said:
> The intention is to follow modelica stdlib

This PR gets that right, since it follows the MSL:
```jl
@connector function Flange(; name)
    sts = @variables begin
        s(t)
        f(t), [connect = Flow]
    end
    ODESystem(Equation[], t, sts, [], name = name, defaults = Dict(s => 0.0, f => 0.0))
end
```
[source](https://github.com/kianmolani/ModelingToolkitStandardLibrary.jl/blob/f7f188e53c6ace0e5adb54a17a777dd0b8be7705/src/Mechanical/Translational/utils.jl#L1-L7)

PR #99, opened 2022-08-19
fixes Symbolics deprecation

ValentinKaisermayer:


ChrisRackauckas:
Bump the symbolics lower bound?

ValentinKaisermayer:
Seems to be an MTK issue with `inputs_to_parameters!`
@baggepinnen

ValentinKaisermayer:
Why doesn't the downstream test fail in MTK?

YingboMa (approved):

ValentinKaisermayer:
> Seems like a real failure?

Yes, but in MTK.

YingboMa (approved):

PR #107, opened 2022-09-14
Start to Mechanical/Translational library

bradcarman:
This is a start to #84 .  Currently this contains simply a Body, Spring, and Damper using a Port for connections


baggepinnen:
See also https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/85

PR #110, opened 2022-09-19
Add analysis points

baggepinnen:
This PR adds something called an `AnalysisPoint`. These are used for linear analyses, such as 
- Computing sensitivity functions.
- Computing complementary sensitivity functions.
- Computing loop-transfer functions.
- Linearizing systems between predefined analysis points.
- Open loops

An `AnalysisPoint` can be created explicitly, but also automatically when `connect`ing an output to an input. Analysis points are causal, and are thus defined in the `Blocks` submodule. They thus expect connections of type `Blocks.RealInput` and `Blocks.ReaOutput`. 

## Notes to reviewers
- ~The current state of the PR does not handle namespaces and subsystems properly. In particular, I'm unsure about the proper way of modifying equations and namespaces when changing the equations in a system with subsystems.~ Namespaces are now handled.
- The user must currently call `expand_analysis_points` manually before simulating a system containing analysis points. It would be good if this would be handled automatically by `structural_simplify`. The current implementation requires `expand_analysis_points` to run before `expand_connections`, but I could not find a clean way of extending this functionality from outside of ModelingToolkit. One approach would be to define `AnalysisPoint` in MTK, but keep the functions related to sensitivity functions here (they require components from MTKStdlib)
- The last commit depends on the PR below
- https://github.com/SciML/ModelingToolkit.jl/pull/1827

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,4 +1,5 @@
 [deps]
+ControlSystemsBase = "aaaaaaaa-a6ca-5380-bf3e-84a91bcd477e"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 IfElse = "615f187c-cbe4-4ef1-ba3b-2fcf58d6d173"
 ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
--- a/test/Mechanical/rotational.jl
+++ b/test/Mechanical/rotational.jl
@@ -60,7 +60,8 @@ end
                                  sine,
                              ])
     sys = structural_simplify(model)
-    prob = DAEProblem(sys, D.(states(sys)) .=> 0.0, [D(D(inertia2.phi)) => 1.0; D.(states(model)) .=> 0.0], (0, 10.0))
+    prob = DAEProblem(sys, D.(states(sys)) .=> 0.0,
+                      [D(D(inertia2.phi)) => 1.0; D.(states(model)) .=> 0.0], (0, 10.0))
     sol = solve(prob, DFBDF())
 
     # Plots.plot(sol; vars=[inertia1.w, -inertia2.w*2])
--- a/Project.toml
+++ b/Project.toml
@@ -19,8 +19,9 @@ Symbolics = "0.1, 1, 2, 3, 4"
 julia = "1.6"
 
 [extras]
+ControlSystemsBase = "aaaaaaaa-a6ca-5380-bf3e-84a91bcd477e"
 SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["SafeTestsets", "Test"]
+test = ["SafeTestsets", "Test", "ControlSystemsBase"]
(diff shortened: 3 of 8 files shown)


test/Mechanical/rotational.jl
```diff
@@ -60,7 +60,8 @@ end
                                  sine,
                              ])
     sys = structural_simplify(model)
-    prob = DAEProblem(sys, D.(states(sys)) .=> 0.0, [D(D(inertia2.phi)) => 1.0; D.(states(model)) .=> 0.0], (0, 10.0))
+    prob = DAEProblem(sys, D.(states(sys)) .=> 0.0,
+                      [D(D(inertia2.phi)) => 1.0; D.(states(model)) .=> 0.0], (0, 10.0))
```
baggepinnen, line 64: Unrelated change made by the formatter

docs/Project.toml
```diff
@@ -1,4 +1,5 @@
 [deps]
+ControlSystemsBase = "aaaaaaaa-a6ca-5380-bf3e-84a91bcd477e"
```
YingboMa, line 2: Remember compat.

baggepinnen: This is in `docs/Project.toml` where no packages have compat so far :P 
https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/de1151f3381b6f4c9a3205c767a076d1627308e6/docs/Project.toml

YingboMa (changes requested):
`flatten` doesn't work with connect, as mentioned in https://github.com/SciML/ModelingToolkit.jl/issues/1826#issuecomment-1255340863 . It should be deprecated and removed. Could you take a look at the `generate_connection_*` functions in MTK? https://github.com/SciML/ModelingToolkit.jl/blob/644d76c73cbeb17458f606c7b299bd450edb0c44/src/systems/connectors.jl Also, it's expensive to traverse the model. It may make sense to add this in MTK. Alternatively, we should provide interfaces that makes expanding connection equations generic.

YingboMa:
This is roughly how connection generation works in MTK.
```julia
julia> sys_without_connections, connection_sets = ModelingToolkit.generate_connection_set(rc_model);

julia> connection_sets
7-element Vector{ModelingToolkit.ConnectionSet}:
 <constant₊output.u(t)::inner, source₊V.u(t)::inner>
 <source₊p.v(t)::inner, resistor₊p.v(t)::inner>
 <source₊p.i(t)::inner, resistor₊p.i(t)::inner>
 <resistor₊n.v(t)::inner, capacitor₊p.v(t)::inner>
 <capacitor₊p.i(t)::inner, resistor₊n.i(t)::inner>
 <capacitor₊n.v(t)::inner, source₊n.v(t)::inner, ground₊g.v(t)::inner>
 <capacitor₊n.i(t)::inner, ground₊g.i(t)::inner, source₊n.i(t)::inner>

julia> connection_equations, stream_equations = ModelingToolkit.generate_connection_equations_and_stream_connections(connection_sets);

julia> connection_equations
8-element Vector{Equation}:
 constant₊output₊u(t) ~ source₊V₊u(t)
 source₊p₊v(t) ~ resistor₊p₊v(t)
 0 ~ resistor₊p₊i(t) + source₊p₊i(t)
 resistor₊n₊v(t) ~ capacitor₊p₊v(t)
 0 ~ capacitor₊p₊i(t) + resistor₊n₊i(t)
 capacitor₊n₊v(t) ~ source₊n₊v(t)
 capacitor₊n₊v(t) ~ ground₊g₊v(t)
 0 ~ capacitor₊n₊i(t) + ground₊g₊i(t) + source₊n₊i(t)
```

PR #113, opened 2022-09-28
Update analysis points to use find/replace in MTK

baggepinnen:
Supersedes #112 
Waiting for 
- https://github.com/SciML/ModelingToolkit.jl/pull/1839


YingboMa (changes requested):
The test is not included.

YingboMa:
Also, there's should be a version bound on MTK.

baggepinnen:
Good catches :) Both should be fixed now

YingboMa (approved):

PR #114, opened 2022-09-29
Start to Translational library with both Velocity and Position based connectors

### LINKED ISSUE #109: Add Mechanical/Translational libraries with both *position* and *velocity* based connectors
Plan:  

- Put the velocity connector approach under "Mechanical/Translational" and position connector approach under "Mechanical/TranslationalPosition" (or maybe "Mechanical/Positional"). 
- Add documentation to explain the theory and detail the differences.  Will serve as a guide to aid in the decision of which approach to use for one's model.

Both libraries will be as similar as possible, with the only difference being the connector across variable.  Because both libraries will export the same names, mixed use can/should only be done with namespaced use. Mixed use should likely be a rare case.  

A start to the documentation can be found here:  [https://github.com/bradcarman/ModelingToolkitStandardLibrary.jl/blob/main/docs/connections.md](https://github.com/bradcarman/ModelingToolkitStandardLibrary.jl/blob/main/docs/connections.md)

bradcarman:
Closes #109


docs/src/connectors/connections.md
```diff
@@ -0,0 +1,354 @@
...
+
+println.(equations(sys))
+nothing # hide
+```
+
+The solution shows what we would expect, a non-linear disipation of voltage and releated decrease in current flow...
+
+```@example connections
+prob = ODEProblem(sys, [1.0], (0, 10.0), [])
+sol = solve(prob, ImplicitMidpoint(); dt=0.01)
```
YingboMa, line 106: In tutorials, we should encourage the user to use the right default. In the case of `solve`, we should use `DifferentialEquations` and use `solve(prob)` only, then `solve` will try to pick the right algorithm automatically. Constant step size implicit midpoint rule is not adequate for most stiff problems because the lack of dumping, albeit well-known.

YingboMa (commented):
Is it intentional to have both `s0` and `s_0`?

bradcarman:
> Is it intentional to have both `s0` and `s_0`?

I see this in the sensor component, I've changed to `s_0` for the next check in

bradcarman:
About connector naming...there is a problem: when defining a block that combines domains, such as a hydrualic actuator that needs hydraulic and mechanical connectors, we will need to be able to distinguish connectors.  For example

```julia
function actuator()

    @named H_a = HydraulicPort()
    @named H_b = HydraulicPort()

    @named MT = TranslationalPort()

    eqs = [
        (H_a.p - H_b.p)*area ~ MT.f

        ...
    ]

end

@named act = actuator()

# when making connections we can easily see what like domains are being connected...
eqs = [
    connect(act.H_a, pump.H)
    connect(mass.MT, act.MT)

]
```

If we decided on a naming scheme now then we can use consistent connector names throughout the Standard Library.

Here is my recommended naming, I am in favor of using capital letters for connectors because they are different from variables, it makes it much easier to read.

```
#Blocks
B
B_a
B_b

#Electrical/Analog
EA
EA_p #positive
EA_n #negative

#Electrical/Digital
ED
ED_p
ED_n

#Magnetic/FluxTubes
MF
MF_a
MF_b

#Mechanical/Translational
MT
MT_a
MT_b

#Mechanical/Rotational
MR
MR_a
MR_b

#Hydraulic
H
H_a
H_b
H_s #source
H_r #return

**Comment by YingboMa** [comment]
What about `delta_s0` vs `delta_s_0`?

**Comment by bradcarman** [comment]
> What about `delta_s0` vs `delta_s_0`?

I think the `_0` will be best to keep a consistent naming like the others `s1_0`,  `s2_0` etc...

**Comment by bradcarman** [comment]
Will stick with `flange`, `flange_a`, etc. for now rather than the connector naming I proposed.  But I wonder, should I change the initial condition parameters `s1_0` to `s_a_0` to better align with `flange_a`?

File: src/Mechanical/Translational/sources.jl [hunk 2]
```diff
@@ -0,0 +1,13 @@
+function SineForce(; name, amp, freq)
```
YingboMa, line 1: `Force` should be generic like the one in TranslationalPosition.

YingboMa (approved):

PR #49, opened 2022-05-08
Adds permanent magnet DC machine 

ValentinKaisermayer:

PR #115, opened 2022-10-04
remove input annotations on RealInput

baggepinnen:
This does not work as intended since the annotated variable resides inside a system without internal connections. Until MTK has more formally specified semantics for causal connections, we rely on the explicit passing of input variables to all functions that require it, such as linearize.

@ValentinKaisermayer 

This PR should resolve https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/49#discussion_r985133712


YingboMa:
We should fix this in MTK instead.

ValentinKaisermayer:
There is already a long lasting issue about this: https://github.com/SciML/ModelingToolkit.jl/issues/1577

PR #118, opened 2022-10-05
Fixes missing plot

ValentinKaisermayer:


baggepinnen:
Wouldn't it be sufficient to remove the nothing on the last line in the code cell?

ValentinKaisermayer:
Maybe, I simply followed the [documentation](https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@example-block).

ChrisRackauckas:
That's a bit silly. Just remove the nothing. No need to save a file and all of that.

PR #134, opened 2022-12-22
more robust detection of analysis-point variable

baggepinnen:
Analysis points are supposed to be used with `RealInput/RealOutput`, and those both have an inner variable called `u`. This PR attempts to connect other kinds of components as well, but throws an error if there is more than one state variable, in which case which one is referred to by the AP is ambiguous.

### DIFF
--- a/src/Blocks/analysis_points.jl
+++ b/src/Blocks/analysis_points.jl
@@ -8,6 +8,13 @@ end
 
 Base.nameof(ap::AnalysisPoint) = ap.name
 
+function ap_var(sys)
+    hasproperty(sys, :u) && return sys.u
+    x = states(sys)
+    length(x) == 1 && return x[1]
+    error("Could not determine the analysis-point variable in system $(nameof(sys)). To use an analysis point, apply it to a connection between two causal blocks containing connectors of type `RealInput/RealOutput` from ModelingToolkitStandardLibrary.Blocks.")
+end
+
 """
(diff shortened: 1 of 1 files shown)


src/Blocks/analysis_points.jl
```diff
@@ -183,15 +190,15 @@ function get_sensitivity_function(sys, ap_name::Symbol; loop_openings = nothing,
             if namespaced_ap_match(ap, ns, ap_name, nothing)
                 namespace[] = ns
                 apr[] = ap
-                (ap.out.u ~ ap.in.u + d), d
+                (ap_var(ap.out) ~ ap_var(ap.in) + d), d
```
YingboMa, line 193: Should we also compare that their variable names equal?

baggepinnen: If there is only a single variable, I figured it was unambiguous no matter what it was called. If there were more than one variable, I threw an error instead. This happens already in the `ap_var` function. 

The reasoning for allowing any one variable is to make it more likely that it will work for user-defined components etc. without the user having built the components using our `RealInput/RealOutput` components.

baggepinnen: Making the matching based on name rather than type is a bit unfortunate in any case. An alternative approach would be to store the variable name in the analysis point, and default to `u` if none was provided. I'm not sure for about how the syntax for this would look though.

YingboMa (approved):

PR #135, opened 2022-12-22
add descriptions to variables and parameters

baggepinnen:
This PR adds `[description = "sting"]` to a number of variables and parameters in the Blocks and Rotational modules. Other modules will come in separate PRs.

This serves two purposes
1. Make models from model libraries easier to work with. Without a description, it's hard to know what a variable called `k` deep inside a model does, in particular if the model comes from a library.
2. Illustrate best practices. People are likely to look to MTKstdlib to learn how to build component libraries, and I hope to establish adding variable descriptions as a best practice people follow.

### DIFF
--- a/src/Blocks/continuous.jl
+++ b/src/Blocks/continuous.jl
@@ -127,16 +127,15 @@ Critical damping corresponds to `d=1`, which yields the fastest step response wi
 function SecondOrder(; name, k = 1, w, d, x_start = 0.0, xd_start = 0.0)
     @named siso = SISO()
     @unpack u, y = siso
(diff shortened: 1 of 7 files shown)


baggepinnen:
JuliaFormatter does not appear to understand the `@variables` macro, when formatting the `continuous.jl` file, it tries to change the syntax 
```julia
xd(t)=xd_start [
        description = "Derivative state of SecondOrder filter $name",
    ]
```

from short to long function definition:
```julia
function xd(t)
        xd_start
    end [
        description = "Derivative state of SecondOrder filter $name",
    ]
```

YingboMa (approved):

PR #144, opened 2023-01-22
Digital: Adds `Logic` and defines 9 logic levels, `StdULogicVector`, and `StdLogicVector`

ven-k:
First in series of PRs to implement the entire IEEE.Std_logic_1164 in Julia

- [x] Adds Logic and its 9 levels
- [x] Accordingly refactors AndTable, OrTable, NotTable, XorTable
- [x] Accordingly refactors _and, _or, _not, _xor
- [x] Adds StdULogicVector and few predefined vectors and StdLogicVector
- [x] Adds relevant tests

----- 
This PR will be followed up to add things like:
- Resolution table
- Inter-conversions from one range of logic-levels to other range
- Digital sources, clocks
- High level DiscreteSystems of gates
- And then revival of Adders, En(/De)coders, (De)Muxes

### DIFF
--- a/src/Electrical/Digital/logic.jl
+++ b/src/Electrical/Digital/logic.jl
@@ -1,0 +1,68 @@
+struct Logic
+    level::Int
+end
+
+function Base.show(io::IO, ::MIME"text/plain", l::Logic)
+    if l.level == 1
+        print(io, "U")
+    elseif l.level == 2
+        print(io, "X")
+    elseif l.level == 3
+        print(io, "F0")
+    elseif l.level == 4
+        print(io, "F1")
+    elseif l.level == 5
+        print(io, "Z")
+    elseif l.level == 6
+        print(io, "W")
+    elseif l.level == 7
+        print(io, "L")
+    elseif l.level == 8
+        print(io, "H")
+    elseif l.level == 9
+        print(io, "DC")
+    else
+        print(io, "Invalid logic level")
+    end
+end
+
+# Aliases for 9 Digital Logic levels
+const Uninitialized = Logic(1)
+const ForcingUnknown = Logic(2)
+const ForcingZero = Logic(3)
+const ForcingOne = Logic(4)
+const HighImpedence = Logic(5)
+const WeakUnknown = Logic(6)
+const WeakZero = Logic(7)
+const WeakOne = Logic(8)
+const DontCare = Logic(9)
+
+const U = Uninitialized
+const X = ForcingUnknown
+const F0 = ForcingZero
+const F1 = ForcingOne
+const Z = HighImpedence
+const W = WeakUnknown
+const L = WeakZero
+const H = WeakOne
+const DC = DontCare
+
+Base.zero(l::Logic) = F0
+Base.zero(l::Type{Logic}) = F0
+Base.one(l::Logic) = F1
+Base.one(l::Type{Logic}) = F1
+
+# Helpers to convert 1 and 0 to their `Logic` counterparts
+function Base.convert(l::Type{Logic}, i::Number)
+    if i == zero(i)
+        zero(l)
+    elseif i == one(i)
+        one(l)
+    else
+        throw("This isn't a valid `Logic` value")
+    end
+end
+Base.convert(l::Type{Logic}, i::Logic) = i
+
+convert_to_logic(i::Number) = convert(Logic, i)
+convert_to_logic(i::Logic) = i
--- a/src/Electrical/Digital/logic.jl
+++ b/src/Electrical/Digital/logic.jl
@@ -30,14 +30,14 @@ function Base.show(io::IO, ::MIME"text/plain", l::Logic)
     elseif Int(l) == 9
         print(io, "DC")
     else
-        print(io, "Invalid logic level")
+        print(io, "Invalid logic level: $l")
     end
 end
 
-Base.zero(l::Logic) = F0
-Base.zero(l::Type{Logic}) = F0
-Base.one(l::Logic) = F1
-Base.one(l::Type{Logic}) = F1
+Base.zero(::Logic) = F0
+Base.zero(::Type{Logic}) = F0
+Base.one(::Logic) = F1
+Base.one(::Type{Logic}) = F1
 
 # Helpers to convert 1 and 0 to their `Logic` counterparts
 function Base.convert(l::Type{Logic}, i::Number)
--- a/src/Electrical/Digital/logic.jl
+++ b/src/Electrical/Digital/logic.jl
@@ -1,41 +1,4 @@
-struct Logic
-    level::Int
-end
-
-function Base.show(io::IO, ::MIME"text/plain", l::Logic)
-    if l.level == 1
-        print(io, "U")
-    elseif l.level == 2
-        print(io, "X")
-    elseif l.level == 3
-        print(io, "F0")
-    elseif l.level == 4
-        print(io, "F1")
-    elseif l.level == 5
-        print(io, "Z")
-    elseif l.level == 6
-        print(io, "W")
-    elseif l.level == 7
-        print(io, "L")
-    elseif l.level == 8
-        print(io, "H")
-    elseif l.level == 9
-        print(io, "DC")
-    else
-        print(io, "Invalid logic level")
-    end
-end
-
-# Aliases for 9 Digital Logic levels
-const Uninitialized = Logic(1)
-const ForcingUnknown = Logic(2)
-const ForcingZero = Logic(3)
-const ForcingOne = Logic(4)
-const HighImpedence = Logic(5)
-const WeakUnknown = Logic(6)
-const WeakZero = Logic(7)
-const WeakOne = Logic(8)
-const DontCare = Logic(9)
+@enum Logic Uninitialized=1 ForcingUnknown ForcingZero ForcingOne HighImpedence WeakUnknown WeakZero WeakOne DontCare
 
 const U = Uninitialized
 const X = ForcingUnknown
@@ -46,6 +9,30 @@ const W = WeakUnknown
 const L = WeakZero
 const H = WeakOne
 const DC = DontCare
+
+function Base.show(io::IO, ::MIME"text/plain", l::Logic)
+    if Int(l) == 1
+        print(io, "U")
+    elseif Int(l) == 2
+        print(io, "X")
+    elseif Int(l) == 3
+        print(io, "F0")
+    elseif Int(l) == 4
+        print(io, "F1")
+    elseif Int(l) == 5
+        print(io, "Z")
+    elseif Int(l) == 6
+        print(io, "W")
+    elseif Int(l) == 7
+        print(io, "L")
+    elseif Int(l) == 8
+        print(io, "H")
+    elseif Int(l) == 9
+        print(io, "DC")
+    else
+        print(io, "Invalid logic level")
+    end
+end
 
 Base.zero(l::Logic) = F0
 Base.zero(l::Type{Logic}) = F0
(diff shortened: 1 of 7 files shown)


src/Electrical/Digital/logic.jl
```diff
@@ -0,0 +1,68 @@
...
+# Aliases for 9 Digital Logic levels
+const Uninitialized = Logic(1)
+const ForcingUnknown = Logic(2)
+const ForcingZero = Logic(3)
+const ForcingOne = Logic(4)
+const HighImpedence = Logic(5)
+const WeakUnknown = Logic(6)
+const WeakZero = Logic(7)
+const WeakOne = Logic(8)
+const DontCare = Logic(9)
```
baggepinnen, line 38: This feels an awful lot like an `@enum`. Have you considered using an enumeration for this type instead?

ven-k: Initially I was considering to include 1 and 0 as is in the set of logic levels; as @enum wouldn't work with that, I dropped the idea then.

Now because of F0, F1 it is doable - Logic as an Enum of these 9 vals.
ATM I can't think of any other field Logic could have. So I will swap this part with an Enum

ven-k: `getfield.(<big StructLogic matrix>, :levels)` vs `Int.(<big EnumLogic matrix>)`
Speeds were very close, and the former was slightly faster. However, the latter always allocated less.

src/Electrical/Electrical.jl
```diff
@@ -26,14 +25,22 @@ include("Analog/sensors.jl")
...
 # TODO:
 # - digital
 # - machines
 # - multi-phase
 
+export Logic
+include("Digital/logic.jl")
+
+export StdLogicVector, StdULogicVector,
+       std_ulogic, UX01, UX01Z, X01, X01Z
```
baggepinnen, line 41: any particular reason for exporting `UX01, UX01Z, X01, X01Z` while not exporting many other vars?

src/Electrical/Digital/logic.jl
```diff
@@ -0,0 +1,57 @@
...
+    elseif Int(l) == 6
+        print(io, "W")
+    elseif Int(l) == 7
+        print(io, "L")
+    elseif Int(l) == 8
+        print(io, "H")
+    elseif Int(l) == 9
+        print(io, "DC")
+    else
+        print(io, "Invalid logic level")
```
baggepinnen, line 33:
```suggestion
        print(io, "Invalid logic level: $l")
```

src/Electrical/Digital/logic.jl
```diff
@@ -0,0 +1,57 @@
...
+    end
+end
+
+Base.zero(l::Logic) = F0
+Base.zero(l::Type{Logic}) = F0
+Base.one(l::Logic) = F1
+Base.one(l::Type{Logic}) = F1
```
baggepinnen, line 40:
```suggestion
Base.zero(::Logic) = F0
Base.zero(::Type{Logic}) = F0
Base.one(::Logic) = F1
Base.one(::Type{Logic}) = F1
```
to avoid potential linter warnings against unused arguments

src/Electrical/Digital/logic.jl
```diff
@@ -0,0 +1,57 @@
...
+    elseif i == one(i)
+        one(l)
+    else
+        throw("$i isn't a valid `Logic` value")
+    end
+end
+Base.convert(l::Type{Logic}, i::Logic) = i
+
+convert_to_logic(i::Number) = convert(Logic, i)
+convert_to_logic(i::Logic) = i
```
baggepinnen, line 55: why are these required when the `Base.convert(l::Type{Logic}, i::Number)` method exists?

src/Electrical/Digital/logic.jl
```diff
@@ -0,0 +1,57 @@
...
+
+get_logic_level(l::Logic) = Int(l)
```
baggepinnen, line 57: Is this function really required when `Int` is equally simple?

src/Electrical/Digital/logic_vectors.jl
```diff
@@ -0,0 +1,47 @@
+import Base: size, axes, getindex
+
+const LogicOrNumber = Union{Logic, Number}
+
+struct StdULogicVector{N} <: AbstractArray{Logic, N}
+    logic::AbstractArray{Logic}
+    level::AbstractArray{Int}
```
baggepinnen, line 7: I notice that there's an abstract type here. I realize that MTK will butcher most type stability anyways and that this may not be a problem in this case, but have the implications of this been thought through?

src/Electrical/Digital/logic_vectors.jl
```diff
@@ -0,0 +1,47 @@
...
+    function StdLogicVector(l::AbstractArray)
+        N = ndims(l)
+        l = AbstractArray{Logic}(convert.(Logic, l))
+        new{N}(l, get_logic_level.(l))
+    end
+end
+
+const LogicVector = Union{StdULogicVector, StdLogicVector}
+
+Base.size(l::L) where {L <: LogicVector} = size(l.logic)
```
baggepinnen, line 27:
```suggestion
Base.size(l::LogicVector) = size(l.logic)
```
this is against the Julia style guide ;)

src/Electrical/Digital/logic_vectors.jl
```diff
@@ -0,0 +1,47 @@
...
+
+Base.axes(l::L) where {L <: LogicVector} = axes(l.logic)
```
baggepinnen, line 29:
```suggestion
Base.axes(l::LogicVector) = axes(l.logic)
```
Same here, the type parameter should really only be introduced if you need to use it (or if you need to force specialization of a function type in some special cases).

src/Electrical/Digital/logic_vectors.jl
baggepinnen: There are several similar instances below. These give precompilation warnings and are best avoided

src/Electrical/Digital/tables.jl
```diff
@@ -1,131 +1,97 @@
-# 9-level logic
-@parameters X
+# AND gate
+AndTable = [
```
baggepinnen, line 2: Should this be `const`?

src/Electrical/Digital/tables.jl
```diff
@@ -1,131 +1,97 @@
...
-    i, j = logic(a), logic(b)
-    (typeof(i) != Int || typeof(j) != Int) && return X
-    OrTable[i, j]
+_not(x::Logic) = NotTable[get_logic_level(x)]
+_not(x::Number) = _not(convert_to_logic(x))
+
+@register_symbolic _not(x)
+
+# OR gate
+OrTable = [
```
baggepinnen, line 40: const?

src/Electrical/Digital/tables.jl
```diff
@@ -1,131 +1,97 @@
...
+_xor2(a::Number, b::Number) = _xor2(convert_to_logic(a), convert_to_logic(b))
 
 function _xor(x...)
-    y = [_xor2(x[1], 0)]
-    for i in 2:length(x)
+    y = [convert_to_logic(x[1])]
+    for i in 2:lastindex(x)
         push!(y, _xor2(x[i], y[i - 1]))
     end
     return y[end]
```
baggepinnen, line 94: You're only ever using the last value of `y`, so do you really need to allocate an array?

src/Electrical/Electrical.jl
ven-k: The IEEE Standard defines these logic vectors.

About the others:
- I expect the logic tables (ex: AndTable) to be used only internally and the logic gates will be exposed to users via (yet-to-be-defined) higher level gate components.
- In the "Standard", Logic values (ex: Uninitialized) are not defined in their full form; but via short form (ex: U, X) and so on. We could export the full form; but usage of short form would be more main stream (; however one could argue that we are deviating with F1, F0 and DC which are simply `1`, `0`, `-`)

src/Electrical/Digital/tables.jl
ven-k: Good point

src/Electrical/Digital/logic.jl
ven-k: Int(<a logic value>) doesn't really clarify what it is returning or what the intent is.

src/Electrical/Digital/logic.jl
ven-k: IMO it would be a convenience function; Could be dropped.

src/Electrical/Digital/logic_vectors.jl
ven-k: Now this is changed to `Array`

baggepinnen (approved):

src/Electrical/Digital/logic_vectors.jl
```diff
@@ -0,0 +1,57 @@
+import Base: size, axes, getindex, setindex!
+
+const LogicOrNumber = Union{Logic, Number}
+
+struct StdULogicVector{N} <: AbstractArray{Logic, N}
+    logic::Array{Logic}
```
YingboMa, line 6: Shouldn't it be `Array{Logic, N}`?

src/Electrical/Digital/logic_vectors.jl
```diff
@@ -0,0 +1,57 @@
...
+
+Base.size(l::LogicVector) = size(l.logic)
+
+Base.axes(l::LogicVector) = axes(l.logic)
+
+function Base.getindex(s::LogicVector,
+                       i::Int)
+    getindex(s.logic, i)
+end
+function Base.getindex(s::LogicVector, i1::Int, i2::Int,
```
YingboMa, line 35: Why do we call it a vector if it can be multidimensional?

src/Electrical/Digital/logic_vectors.jl
```diff
@@ -0,0 +1,57 @@
+import Base: size, axes, getindex, setindex!
+
+const LogicOrNumber = Union{Logic, Number}
+
+struct StdULogicVector{N} <: AbstractArray{Logic, N}
+    logic::Array{Logic}
+    level::Array{Int}
```
YingboMa, line 7: Why do we need to store `level` at all if it can be trivially computed from `logic`?

src/Electrical/Digital/logic_vectors.jl
```diff
@@ -0,0 +1,57 @@
...
+    function StdLogicVector(l::Array)
+        N = ndims(l)
+        l = Array{Logic}(convert.(Logic, l))
+        new{N}(l, get_logic_level.(l))
+    end
+end
+
+const LogicVector = Union{StdULogicVector, StdLogicVector}
```
YingboMa, line 25: What are they used for?

src/Electrical/Digital/logic_vectors.jl
ven-k: To remain consistent with `std_logic_vector` and `std_ulogic_vector` are the names used in IEEE Std_logic_1164.
Although its mostly used in vector form, it can be a matrix as well.

src/Electrical/Digital/logic_vectors.jl
YingboMa: Should we have a logic table so that it takes `Logic` indices, then?

src/Electrical/Digital/logic_vectors.jl
ven-k: StdULogicVector, StdLogicVector are part the Std_logic_1164.
LogicVector is only for convenience while writing dispatches.

src/Electrical/Digital/logic_vectors.jl
ven-k: Done

src/Electrical/Digital/logic_vectors.jl
YingboMa: This point is still not addressed. We need `LogicTable{N} <: AbstractArray{Logic, N}` such that we can index into it using `Logic` indices, so we don't even need to store `level`.

ven-k: This seemed like the simplest approach to index the values.
I see the point in a `LogicTable` type. Will implement that.

YingboMa (approved):

PR #145, opened 2023-01-26
add `TimeVaryingFunction` source block

baggepinnen:
This is useful to input data from, e.g., an interpolator


baggepinnen:
I'm not sure what's up with the format check, the formatter is happy locally.

ChrisRackauckas:
Are you on the right version of formatter? It recently updated the format.

PR #146, opened 2023-01-27
allow the user to override the default smoothing parameter for sources

baggepinnen:


ven-k:
Changes LGTM.

But wouldn't it be better if
1. We have dispatches based on smooth - Int or Bool?
Or
2. Just have another var `δ` or `smooth_coef``

baggepinnen:
1. You cannot dispatch on keyword arguments, and the workaround to do so makes the code very bloated
2. I didn't want two different arguments to handle the same thing, both of which must be set to achieve a change in smoothness factor
3. I don't want to break the existing interface. 

I think this solution was a good tradeoff

ven-k (approved):

jebej:
Doesn’t this change make those functions type unstable when `smooth::Bool`?

baggepinnen:
They do, but the output is a symbolical expression, which is type unstable anyways so what does it matter?

PR #163, opened 2023-05-01
Input component and Hydraulic changes

bradcarman:
- Input component that can now properly drive models from data.  See: input component tutorial for more information
- Hydraulic/IsothermalCompressible now exports Tube instead of Pipe so as to not conflict with Base.Pipe
- Added Actuator component to Hydraulic library

### DIFF
--- a/src/Blocks/sources.jl
+++ b/src/Blocks/sources.jl
@@ -412,7 +412,7 @@ end
 # - SawTooth    Generate saw tooth signal
 # - Trapezoid   Generate trapezoidal signal of type Real
 
-function linear_interpolation(x1::T, x2::T, t1::T, t2::T, t::T) where {T<:Real}
+function linear_interpolation(x1::T, x2::T, t1::T, t2::T, t::T) where {T <: Real}
     if t1 != t2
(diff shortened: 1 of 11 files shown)


YingboMa (approved):

src/Blocks/sources.jl
```diff
@@ -411,3 +411,160 @@ end
...
+        t = zero(t)
+    end
+
+    if isempty(memory.data)
+        if T isa Float16
+            return NaN16
+        elseif T isa Float32
+            return NaN32
+        elseif T isa Float64
+            return NaN64
```
baggepinnen, line 502:
```julia
T(NaN)
```

baggepinnen (commented):
It's not completely clear to me what kind of input data this new `Input` component can handle, does the data have to be equidistantly sampled? I think the name `Input` might be slightly unfortunate since
1. `input` is a keyword in modelica with a completely different meaning. We have a similar metadata field in MTK `[input = true]` which also has a different semantic meaning.
2. The functionality behind the `Input` component appears to be limited to sampled data inputs only, and the sample rate must be equidistant? `SampledDataInput` may be a more descriptive name.

PR #174, opened 2023-05-19
Fixes to test suite

staticfloat:
This PR contains a set of fixes to the test suite that should be always desirable.  In a separate PR, I have collected fixes that may not be desirable, but are required for running with DAECompiler.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -18,9 +18,10 @@ julia = "1.6"
 [extras]
 ControlSystemsBase = "aaaaaaaa-a6ca-5380-bf3e-84a91bcd477e"
 DataInterpolations = "82cc6244-b520-54b8-b5a6-8a565e85f1d0"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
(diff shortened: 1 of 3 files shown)


test/Thermal/thermal.jl
```diff
@@ -32,7 +32,7 @@ D = Differential(t)
     # Check if Relative temperature sensor reads the temperature of heat capacitor
     # when connected to a thermal conductor and a fixed temperature source
     @test sol.retcode == Success
-    @test sol[reltem_sensor.T] + sol[tem_src.port.T] == sol[mass1.T] + sol[th_conductor.dT]
+    @test sol[reltem_sensor.T] + sol[tem_src.port.T] ≈ sol[mass1.T] + sol[th_conductor.dT]
```
YingboMa, line 35: I am not sure about these ones. Shouldn't derived variables be exact?

staticfloat: DAECompiler is off by a few ulps, I'm guessing because state tearing chose different states than MTK did, causing the reconstruction of these values to perform floating-point operations in a different order.  I haven't looked deeply into it, I just saw that the answer was off by a few ulps so I changed it to `≈`.

staticfloat:
I've dropped the `\approx` changes, I will open a separate PR for those.

YingboMa (approved):

PR #180, opened 2023-05-30
Optimize `friction_factor` function

bradcarman:


YingboMa (approved):
This is a good trick to fix the instability

PR #188, opened 2023-06-26
fix: Ground with new model syntax

jvaverka:
Description:
Convert all components needed for [DC Motor with PI-controller](https://docs.sciml.ai/ModelingToolkitStandardLibrary/stable/tutorials/dc_motor_pi/#DC-Motor-with-PI-controller) to the new `@model` syntax.

Questions:
- When should `@extend` be used in favor of `@components`?

### DIFF
--- a/src/Electrical/Analog/ideal_components.jl
+++ b/src/Electrical/Analog/ideal_components.jl
@@ -8,10 +8,9 @@ node.
 
   - `g`
 """
-@component function Ground(; name)
-    @named g = Pin()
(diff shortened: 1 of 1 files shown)


jvaverka:
@baggepinnen
@bauglir
@ven-k

baggepinnen:
> When should @extend be used in favor of @components?

@jvaverka I go through the differences between compose (which is equivalent to providing `systems = ...`) and `extend` in this video https://youtu.be/eLNX9-znzU0?t=3284

PR #200, opened 2023-07-20
Don't export IdealGear from TranslationalPosition and TranslationalModelica

ven-k:
- It isn't implemented in those libs

I don't see any implementation of IdealGear in these libs. @baggepinnen can we just drop them or do we want to reexport Rotational.IdealGear?


baggepinnen:
The export probably came with copy-paste, let's remove the export until a translational gear has been implemented :+1:

baggepinnen (approved):

ven-k:
cc @ChrisRackauckas

PR #209, opened 2023-08-14
refactor: kwargs of base system are now kwargs of main component

ven-k:
- With this https://github.com/SciML/ModelingToolkit.jl/pull/2226, kwargs of base system as kwargs of main sys.
- Symbolic defaults can be set. Hence `use_support` is by default `false` everywhere.
- Tests are updated accordingly.
- Depends on https://github.com/SciML/ModelingToolkit.jl/pull/2226


YingboMa (commented):
We should add a bound on MTK as well.

YingboMa (approved):

ven-k:
In the invalidations test, this branch successfully precompiles- see[^1].
However it is the default branch (against which invalidations count is taken), fails to precompile. That should be expected with  MTK@8.67 and this PR is meant to fix the errors like:

```julia
ERROR: LoadError: syntax: function argument name not unique: "y_start" around /home/runner/work/ModelingToolkitStandardLibrary.jl/ModelingToolkitStandardLibrary.jl/src/Blocks/nonlinear.jl:77
```

So this PR can be merged.

[^1]:https://github.com/SciML/ModelingToolkitStandardLibrary.jl/actions/runs/6079874565/job/16492986525#step:5:292

PR #212, opened 2023-08-21
Use lowest order differentiated variables to improve state selection

YingboMa:
Note that we can infer `D(x) = D(y)` from `x = y`, but we cannot infer `x = y` from `D(x) = D(y)`, so using the lowest variable in the connector provides more information to the compiler.


YingboMa:
I will update MTK test once this PR is merged.

PR #221, opened 2023-09-25
avoid using array variable

baggepinnen:
since they lead to `p::Union{Float64, Matrix{Float64}}`. Iäve tried a few different ways of `collect`ing the array variable without success, so this PR simply changes the array to a structural parameter to get something that is at least working


baggepinnen:
Test failure is unrelated

ven-k (approved):
LGTM

PR #229, opened 2023-10-03
initialize all relevant variables for SISO

baggepinnen:
This PR makes sure to pass initial values to also connector variables (they are supposed to be equal), and by default initializes the limiter to the center of the limit band. Having the limiter in the non-saturated region by default is important for linearization, you get the zero system otherwise.


ven-k (approved):
It was separately set to 0.0 as defaults didn't work back then. And `u_start` and `y_start` were added to avoid breaking the API when this update is added. LGTM.

PR #230, opened 2023-10-03
improve docstring and interface to PI controller

baggepinnen:
The previous interface required the user to provide a value of `gainPI.k`, now there is an outer parameter `k` that forwards to the inner parameter. This makes `PI` consistent with the other controller definitions and avoids mentioning the internal implementation detail `gainPI`.

### DIFF
--- a/src/Blocks/continuous.jl
+++ b/src/Blocks/continuous.jl
@@ -172,14 +172,19 @@ Initial value of the state `x` can be set with `x`, and of derivative state `xd`
 end
 
 """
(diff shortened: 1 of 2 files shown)


ven-k:
Here: https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/230/files#diff-2c628e949fda2edfd7af3ee31fbf986535a6b1e959d92d03be9ca070a4b633aeR209
it should be changed to `gainPI = Gain(; k = k)`

ven-k (approved):

PR #231, opened 2023-10-03
Fix for PID component to accept symbolic parameters

### LINKED ISSUE #203: PID in Blocks not able to accept symbolic parameters
The following code gives `ERROR: TypeError: non-boolean (Num) used in boolean context`

```julia
using ModelingToolkit
using ModelingToolkitStandardLibrary.Blocks

@parameters k = 3 Ti = 0.5 Td = 1 / 100
@named pid_controller = PID(;k, Ti, Td)
```

bradcarman:
fixes #203

### DIFF
--- a/src/Blocks/Blocks.jl
+++ b/src/Blocks/Blocks.jl
@@ -26,7 +26,7 @@ export Limiter, DeadZone, SlewRateLimiter
 include("nonlinear.jl")
 
 export Integrator, Derivative, FirstOrder, SecondOrder, StateSpace
-export PI, LimPI, PID, LimPID
+export PI, LimPI, PID, PD, LimPID
 include("continuous.jl")
 
 export AnalysisPoint, get_sensitivity, get_comp_sensitivity,
--- a/src/Blocks/continuous.jl
+++ b/src/Blocks/continuous.jl
@@ -214,15 +214,15 @@ See also [`LimPI`](@ref)
 end
 
 """
-    PID(;name, k=1, Ti=false, Td=false, Nd=10, int__x=0, der__x=0)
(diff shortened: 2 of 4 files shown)


baggepinnen (commented):
Thanks for fixing with the parameters :)

I'm not sure if I like the new syntax though
```julia
@named pid_controller = PID("PI"; k = 3, Ti)
```
If you want a PI controller, why don't you call `PI` instead of `PID`? Using strings as arguments to select options is rather uncommon in Julia.

This is also a breaking change since you have changed the method to turn `I` and `D` off from passing `false`  for their respective parameters to passing `with_I` etc., and I am strictly against breaking changes unless they are extremely well motivated _and_ accompanied by a major version bump (which itself is associated with a large amount of extra work for everyone).

bradcarman:
I see this has unfortunately breaking changes.  However, this is something that needs to get fixed.  Currently it's not possible to build a controller system that owns the parameters that are fed to the PID since currently the inputs for Ti and Td need to be strictly numeric values and not @parameters.  I'm not a fan of the proposed interface, but it's the best I could come up with since we are forced to separate `with_I` and `with_D` from the actual inputs of `Ti` and `Td`.  I wish there was a better way, maybe a new ModelingToolkit.jl feature could allow for the original interface.

baggepinnen:
Maybe we could avoid breaking the interface by

1. Letting `PID` work exactly the same if the provided parameters are numeric
2. Have a separate `PD` component to handle the case where there is no integrator. We already have `PI`.
3. Let `PID` always be `PID` if parameters are symbolic. 

That way we would never need any structural parameters, since the structure is determined by the user selecting the appropriate component out of `PI, PD, PID`.

Would this solve your problem?

bradcarman:
Yeah, I think this is the right way to go.

PR #232, opened 2023-10-17
nicify some docstrings and descriptions

baggepinnen:
After playing around with the GUI a bit, I realized that some descriptions were overly verbose and at the same time not as informative as they could be.

### DIFF
--- a/src/Blocks/continuous.jl
+++ b/src/Blocks/continuous.jl
@@ -1,8 +1,8 @@
 """
     Integrator(;name, k = 1, x = 0.0)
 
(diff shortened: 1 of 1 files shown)


ven-k:
There were some discussions on whether we should add `label` metadata and use that for one word description like `Resistance`, `Gain` etc, while `description` can be long form description.

baggepinnen:
Yeah, that could be useful. The descriptions that were shortened were so because they didn't actually provide anything meaningful with all their words: "Time constant of FirstOrder filter" is no more informative than "Time constant" since it's obvious from context, both in the GUI and in the REPL, that it belongs to the FirstORder filter

PR #214, opened 2023-08-28
Add units and missing descriptions to components

ven-k:
Closes https://github.com/SciML/ModelingToolkitStandardLibrary.jl/issues/207

PR #240, opened 2023-11-08
docs: add cauer low pass analog filter example

jvaverka:


docs/src/tutorials/cauer_low_pass_analog.md
```diff
@@ -0,0 +1,250 @@
...
+```
+
+What is happening with `ModelingToolkit.missing_variable_defaults(model)`?
+If you try and run `ODEProblem(model, (0.0, 60.0))` then you will encounter the error:
+
+```plaintext
+ERROR: ArgumentError: Equations (7), states (7), and initial conditions (2) are of different lengths.
+```
+
+This occurs because dummy derivatives where generated without defined initial values.
```
baggepinnen, line 224:
```suggestion
This occurs because dummy derivatives were generated without defined initial values.
```

docs/src/tutorials/cauer_low_pass_analog.md
```diff
@@ -0,0 +1,250 @@
+# Cauer Low Pass Analog Filter
+
+The example Cauer Filter is a low-pass-filter of the fifth order.
```
baggepinnen, line 3:
```suggestion
The example Cauer Filter is a low-pass filter of the fifth order.
```

docs/src/tutorials/cauer_low_pass_analog.md
```diff
@@ -0,0 +1,250 @@
...
+It is realized using an analog network.
+The [`Electrical.Voltage`](@ref) source is the input voltage (whose value varies as defined by [`Blocks.Step`](@ref)), and the `resistor.p.v` is the filter output voltage.
+The pulse response is calculated.
```
baggepinnen, line 6: Do you mean _step_ response?

baggepinnen (commented):
Nice, thanks for this detailed tutorial :)

Would it be useful to add units to the capacitance and inductance parameters? 
@ven-k has been adding units to components in
- https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/214
I'm not sure if adding units in this tutorial will work before that PR is merged?

docs/src/tutorials/cauer_low_pass_analog.md
```diff
@@ -0,0 +1,250 @@
...
+See [ModelingToolkit FAQ](https://docs.sciml.ai/ModelingToolkit/stable/basics/FAQ/) for more information on this and other common questions.
+
+### Solving the Problem
+
+You are finally ready to solve!
+However, if you try to run `solve(prob)` then you will encounter an error.
+
+```plaintext
+ERROR: Default algorithm choices require DifferentialEquations.jl.
+Please specify an algorithm (e.g., `solve(prob, Tsit5())` or
```
ven-k, line 235: Perhaps we can suggest `using DifferentialEquations` and `solve(prob)` instead of OrdinaryDiffEq? 
(While keeping the note+url on where they can find all the solvers they can specify if they want to.)

ven-k:
I'm keeping the Units PR open till upcoming breaking release; it would invalidate any model that use these along with unitless parameters.

docs/src/tutorials/cauer_low_pass_analog.md
jvaverka:
```suggestion
The example Cauer Filter is a low-pass-filter of the fifth order.
It is realized using an analog network.
The [`Electrical.Voltage`](@ref) source is the input voltage (whose value varies as defined by [`Blocks.Step`](@ref)), and the `resistor.p.v` is the filter output voltage.
The step response is calculated.
```

Yes, that is correct.

docs/src/tutorials/cauer_low_pass_analog.md
jvaverka: You mean for all the tutorials? Seems like a good idea to me. My understanding of the trade-off being `OrdinaryDiffEq.jl` is more lightweight and therefore faster to load, but `DifferentialEquations.jl` is more robust and offers conveniences to new users - which is useful for readers going through the tutorials.

Should I leave the note, and open an issue to switch over all tutorials and update the `Project.toml`?

docs/src/tutorials/cauer_low_pass_analog.md
jvaverka: https://github.com/SciML/ModelingToolkitStandardLibrary.jl/commit/7a90bc4d8b094fa13fd91a02b540f9ae22f8e733

PR #257, opened 2024-02-02
docs: add translational components in the API section

sathvikbhagavan:


ChrisRackauckas:
Looks like this has docs build failures?

sathvikbhagavan:
The doc build passes now. It was failing because things like https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/main/src/Mechanical/Translational/components.jl#L49 where it treats `[m] (optional)` in the docstring as a link which is malformed and hence errors. Have to be careful around that.

sathvikbhagavan:
@ChrisRackauckas, can this be merged?

PR #261, opened 2024-02-05
Define `isconnection` on AnalysisPoint

YingboMa:
This change is required for https://github.com/SciML/ModelingToolkit.jl/pull/2441


YingboMa:
This PR should be a part of the MTK V9 update.

PR #274, opened 2024-03-10
Rotational: default values are now guesses + fix tests

ven-k:
- Defaults in rotational lib are `guess`es
- Passes du0 and u0 for tests (in Rotational, Blocks)
- Fixes the linear analysis tests and docs (related to DC Motor example)

## Checklist

### DIFF
--- a/src/Blocks/continuous.jl
+++ b/src/Blocks/continuous.jl
@@ -15,12 +15,12 @@ Initial value of integrator state ``x`` can be set with `x`
 
 # Unknowns:
 
-  - `x`: State of Integrator (with initial guess of 0.0)
+  - `x`: State of Integrator. Defaults to 0.0.
 """
 @mtkmodel Integrator begin
     @extend u, y = siso = SISO()
     @variables begin
-        x(t), [description = "State of Integrator", guess = 0.0]
+        x(t) = 0.0, [description = "State of Integrator"]
     end
     @parameters begin
         k = 1, [description = "Gain"]
@@ -56,7 +56,7 @@ A smaller `T` leads to a more ideal approximation of the derivative.
 
 # Unknowns:
 
-  - `x`: Unknown of Derivative (with initial guess of 0.0)
+  - `x`: Unknown of Derivative. Defaults to 0.0.
 
 # Connectors:
 
@@ -66,7 +66,7 @@ A smaller `T` leads to a more ideal approximation of the derivative.
 @mtkmodel Derivative begin
     @extend u, y = siso = SISO()
     @variables begin
-        x(t), [description = "Derivative-filter state", guess = 0.0]
+        x(t) = 0.0, [description = "Derivative-filter state"]
     end
     @parameters begin
         T = T, [description = "Time constant"]
--- a/src/Blocks/continuous.jl
+++ b/src/Blocks/continuous.jl
@@ -15,12 +15,12 @@ Initial value of integrator state ``x`` can be set with `x`
 
 # Unknowns:
 
-  - `x`: Unknown of Integrator (with initial guess of 0.0)
+  - `x`: State of Integrator (with initial guess of 0.0)
 """
 @mtkmodel Integrator begin
     @extend u, y = siso = SISO()
     @variables begin
-        x(t), [description = "Unknown of Integrator", guess = 0.0]
+        x(t), [description = "State of Integrator", guess = 0.0]
     end
     @parameters begin
         k = 1, [description = "Gain"]
@@ -30,6 +30,7 @@ Initial value of integrator state ``x`` can be set with `x`
         y ~ x
     end
 end
+
 """
     Derivative(; name, k = 1, T, x = 0.0)
 
--- a/src/Blocks/continuous.jl
+++ b/src/Blocks/continuous.jl
@@ -12,11 +12,15 @@ Initial value of integrator state ``x`` can be set with `x`
 # Parameters:
 
   - `k`: Gain of integrator
+
+# Unknowns:
+
+  - `x`: Unknown of Integrator (with inital guess of 0.0)
 """
 @mtkmodel Integrator begin
     @extend u, y = siso = SISO()
     @variables begin
-        x(t) = 0.0, [description = "State of Integrator"]
+        x(t), [description = "Unknown of Integrator", guess = 0.0]
     end
     @parameters begin
         k = 1, [description = "Gain"]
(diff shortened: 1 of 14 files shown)


test/Mechanical/rotational.jl
```diff
@@ -8,215 +8,219 @@ using OrdinaryDiffEq: ReturnCode.Success
...
+        end
+    end
 
-    @named model = ODESystem(connections, t,
-        systems = [fixed, inertia1, inertia2, spring, damper])
-    sys = structural_simplify(model)
+    @mtkbuild sys = TwoInertia()
 
-    prob = ODEProblem(sys, Pair[], (0, 10.0))
+    prob = ODEProblem(sys, [D(D(sys.inertia2.phi)) => 0], (0, 10.0))
```
ven-k, line 28: @YingboMa this and in following ODE and DAE problems, now we have to pass an initial value to `sys₊inertia2₊phi_tt`. However, while adding this lib this was discussed and after some changes in tearing (in MTK), this was no longer required.
Why is it required again?

ven-k:
@ChrisRackauckas this is ready for review.

src/Blocks/continuous.jl
```diff
@@ -12,11 +12,15 @@ Initial value of integrator state ``x`` can be set with `x`
...
+
+# Unknowns:
+
+  - `x`: Unknown of Integrator (with initial guess of 0.0)
 """
 @mtkmodel Integrator begin
     @extend u, y = siso = SISO()
     @variables begin
-        x(t) = 0.0, [description = "State of Integrator"]
+        x(t), [description = "Unknown of Integrator", guess = 0.0]
```
baggepinnen, line 23: This instance of "state" should not be translated to "unknown", state is indeed correct here

test/Blocks/test_analysis_points.jl
```diff
@@ -165,7 +166,7 @@ function SystemModel(u = nothing; name = :model)
            connect(inertia2.flange_a, spring.flange_b, damper.flange_b)]
     if u !== nothing
         push!(eqs, connect(torque.tau, u.output))
-        return @named model = ODESystem(eqs, t;
+        return ODESystem(eqs, t;
```
ChrisRackauckas, line 169: then the name needs to be passed

src/Mechanical/Rotational/components.jl
```diff
@@ -167,7 +167,7 @@ Linear 1D rotational spring and damper
     @parameters begin
         d, [description = "Damping constant"]
         c, [description = "Spring constant"]
-        phi_rel0 = 0.0, [description = "Unstretched spring angle"]
+        phi_rel0, [description = "Unstretched spring angle", guess = 0.0]
```
ven-k, line 170: This is an oversight; will revert.

src/Blocks/continuous.jl
ven-k: Resolved here: https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/274/commits/eae592ba86e9de11fa575f005d3c13b526a80c99#diff-2c628e949fda2edfd7af3ee31fbf986535a6b1e959d92d03be9ca070a4b633aeR23

src/Mechanical/Rotational/components.jl
ven-k: Resolved here: https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/274/commits/eae592ba86e9de11fa575f005d3c13b526a80c99#diff-36a0d1920684771c0166099143e51f1cf58f1faae7d847cb151c0a0826c4348bR170

test/Blocks/test_analysis_points.jl
ven-k: Resolved here: https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/274/commits/eae592ba86e9de11fa575f005d3c13b526a80c99#diff-3603e44df7347f59dcb8354b6aa1af55ec51d86974ae7a5521ae7214cba54c27R177-R178

ven-k:
Downgrade CI doesn't seem to be happy yet; but I will recheck up to which versions I can bump them to (as a follow up). I don't want to increase the compats to latest available versions; but keep them just above the minimum required ones.

ven-k:
The failing Downgrade CI test is fixed by:
- https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/276

baggepinnen:
Considering how much of the user-facing documentation and tests had to be changed in this PR, I'd say this is a massively breaking change that requires a new major version?

src/Blocks/continuous.jl
```diff
@@ -12,11 +12,15 @@ Initial value of integrator state ``x`` can be set with `x`
...
+
+# Unknowns:
+
+  - `x`: State of Integrator (with initial guess of 0.0)
 """
 @mtkmodel Integrator begin
     @extend u, y = siso = SISO()
     @variables begin
-        x(t) = 0.0, [description = "State of Integrator"]
+        x(t), [description = "State of Integrator", guess = 0.0]
```
baggepinnen, line 23: There is no guess needed here, this should be a default value.

src/Blocks/continuous.jl
```diff
@@ -57,7 +66,7 @@ A smaller `T` leads to a more ideal approximation of the derivative.
 @mtkmodel Derivative begin
     @extend u, y = siso = SISO()
     @variables begin
-        x(t) = 0.0, [description = "Derivative-filter state"]
+        x(t), [description = "Derivative-filter state", guess = 0.0]
```
baggepinnen, line 69: Same here, there is no guessing involved, a default is much better.

ven-k:
> I'd say this is a massively breaking change that requires a new major version?

This sounds good. Even the Units PR will be unblocked with this PR. So I'd say we can have that in too.
(And probably guesses in other sub-libraries as well - with dedicated PRs for each libraries)

> Was there a particular reasons these two blocks and only these were changed?

The idea is - incrementally change all the sub-libraries; tend to the ones with broken tests first.

The analysis-point tests were failing as the rotational components now have guesses instead of defaults - fixed with `op` - while there, I changed these two as they were involved.

I've reverted the changes to `Integrator` and `Derivative` blocks.

src/Blocks/continuous.jl
ven-k: Resolved with: https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/274/commits/3a7ccfbd43de07b1f1de3f81dc1b86e173146eac

src/Blocks/continuous.jl
ven-k: Resolved with: https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/274/commits/3a7ccfbd43de07b1f1de3f81dc1b86e173146eac

baggepinnen:
This was released in a minor version but it should have been a major version bump 😔

ChrisRackauckas:
It probably should have.

PR #283, opened 2024-04-12
refactor(Blocks): add `RealInputArray` and `RealOutputArray`

sathvikbhagavan:
This adds `RealInputArray` and `RealOutputArray` components
This also changes the default values of the states in the RealInput/Output/Array components to guesses.

This is useful for using symbolic vector for input/output size to be 1. Needed for using Neural Network blocks. Ref - https://github.com/SciML/ModelingToolkitNeuralNets.jl/issues/16

### DIFF
--- a/src/Blocks/utils.jl
+++ b/src/Blocks/utils.jl
@@ -1,5 +1,6 @@
-@connector function RealInput(; name, nin = 1, u_start = nin > 1 ? zeros(nin) : 0.0)
-    if nin == 1
+@connector function RealInput(;
+        name, nin = 1, isarray = false, u_start = isarray ? zeros(nin) : 0.0)
+    if nin == 1 && !isarray
         @variables u(t)=u_start [
             input = true,
             description = "Inner variable in RealInput $name"
@@ -21,13 +22,15 @@ Connector with one input signal of type Real.
 # Parameters:
 - `nin=1`: Number of inputs
 - `u_start=0`: Initial value for `u`
+- `isarray=false`: This is only applicable for `nin=1`. Boolean flag to use a scalar or a one element symbolic vector
 
 # States:
 - `u`: Value of the connector; if nin=1 this is a scalar
 """ RealInput
 
-@connector function RealOutput(; name, nout = 1, u_start = nout > 1 ? zeros(nout) : 0.0)
-    if nout == 1
+@connector function RealOutput(;
+        name, nout = 1, isarray = false, u_start = isarray ? zeros(nout) : 0.0)
+    if nout == 1 && !isarray
         @variables u(t)=u_start [
             output = true,
             description = "Inner variable in RealOutput $name"
@@ -49,6 +52,7 @@ Connector with one output signal of type Real.
 # Parameters:
 - `nout=1`: Number of outputs
 - `u_start=0`: Initial value for `u`
+- `isarray=false`: This is only applicable for `nout=1`. Boolean flag to use a scalar or a one element symbolic vector
 
 # States:
 - `u`: Value of the connector; if nout=1 this is a scalar
--- a/src/Blocks/utils.jl
+++ b/src/Blocks/utils.jl
@@ -1,6 +1,6 @@
-@connector function RealInput(;
-        name, nin = 1, isarray = false, u_start = isarray ? zeros(nin) : 0.0)
-    if nin == 1 && !isarray
+@connector function RealInput(; name, nin = 1, u_start = nin > 1 ? zeros(nin) : 0.0)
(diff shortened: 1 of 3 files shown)


YingboMa (changes requested):
@ChrisRackauckas and I believe that this is not correct in general. `RealInput` should be for just scalars and we should add a separate `RealInputArray` component instead. For backward compat, we should keep `nin`, etc in `RealInput` and add a warning when `nin>1`.

sathvikbhagavan:
@YingboMa @ChrisRackauckas, I have added the components for `RealInputArray` and `RealOutputArray`.

src/Blocks/utils.jl
```diff
@@ -42,18 +64,38 @@ Connector with one input signal of type Real.
...
+    u = collect(u)
+    ODESystem(Equation[], t, [u...], []; name = name)
+end
+@doc """
+    RealOutputArray(;name, nout, u_start)
+
+Connector with an array of output signals of type Real.
+
+# Parameters:
+- `nout=2`: Number of outputs.
```
ChrisRackauckas, line 92: That default makes no sense. Why `2`?

src/Blocks/utils.jl
```diff
@@ -14,19 +15,40 @@
...
 @doc """
-    RealInput(;name, nin, u_start)
+    RealInput(;name, u_start)
 
 Connector with one input signal of type Real.
 
 # Parameters:
-- `nin=1`: Number of inputs
-- `u_start=0`: Initial value for `u`
+- `u_start=0`: Initial value for `u`.
```
ChrisRackauckas, line 23: Why does this have an initial value default?

sathvikbhagavan: Shouldn't there be a default value for the variable?

```julia
julia> @named x = RealInput(; nin = 1, u_start = 1.0)
Model x with 0 (1) equations
Unknowns (1):
  u(t) [defaults to 1.0]: Inner variable in RealInput x
Parameters (0):
```

src/Blocks/utils.jl
sathvikbhagavan: As this was an array, I thought 2 element might be more common than 1 element for the default. I can change it to 1 if thats better.

src/Blocks/utils.jl
ChrisRackauckas: Won't that cause issues in initialization?

src/Blocks/utils.jl
ChrisRackauckas: Why would it have a default at all? This seems like it should be a required argument

sathvikbhagavan: Yes, that would be better - addressed in [6bb905a](https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/283/commits/6bb905a45acfbdbf2517372ee99697e3c6de236a)

src/Blocks/utils.jl
sathvikbhagavan: Addressed in [4cfc041](https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/283/commits/4cfc041d9f6d242c48e5a5df41758d89ce23eb25)

sathvikbhagavan: Removing the default is erroring - https://github.com/SciML/ModelingToolkitStandardLibrary.jl/actions/runs/8706828508/job/23880355450?pr=283#step:6:577 even though there is an initial guess 🤔

sathvikbhagavan: After talking with Aayush, we found guesses are not propagated in a hierarchical system. Hopefully https://github.com/SciML/ModelingToolkit.jl/pull/2654 should fix it

sathvikbhagavan: One more bug from MTK - `linearize_function` doesn't use guesses - https://github.com/SciML/ModelingToolkit.jl/pull/2656 should fix it

ChrisRackauckas: Linearization shouldn't necessarily use guesses. It should make sure the algebraic conditions are satisfied and that's it. That's a separate issue though, just update the test to be good and we can merge, the linearization failure is separate from this.

sathvikbhagavan: Do you mean to mark it as broken/skip? done that in [9bc4b8b](https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/283/commits/9bc4b8bceef58149cd00932bba791b54d208ae0b)

ChrisRackauckas: no, fix it by passing in state values

PR #291, opened 2024-04-29
fix: fix guesses for `Real*Input` and `Real*Output`

AayushSabharwal:
## Checklist

- [ ] Appropriate tests were added
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the

### DIFF
--- a/src/Blocks/utils.jl
+++ b/src/Blocks/utils.jl
@@ -1,5 +1,11 @@
-@connector function RealInput(; name, nin = 1, u_start = nin > 1 ? zeros(nin) : 0.0)
+@connector function RealInput(;
+        name, nin = 1, u_start = nothing, guess = nin > 1 ? zeros(nin) : 0.0)
     nin > 1 && @warn "For inputs greater than one, use `RealInputArray`."
+    if u_start !== nothing
+        Base.depwarn(
+            "The keyword argument `u_start` is deprecated. Use `guess` instead.", :u_start)
+        guess = u_start
+    end
     if nin == 1
         @variables u(t) [
             input = true,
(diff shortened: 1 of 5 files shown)


src/Blocks/utils.jl
```diff
@@ -10,9 +10,8 @@
             input = true,
             description = "Inner variable in RealInput $name"
         ]
-        u = collect(u)
     end
-    ODESystem(Equation[], t, [u...], []; name = name, guesses = [u => u_start])
+    ODESystem(Equation[], t, [u], []; name = name, guesses = [u => u_start])
```
baggepinnen, line 14: If I as a user pass `u_start=1` I want that to be true, not only a guess. I wonder if these `u_start` are needed at all?

ChrisRackauckas: You should be able to change the guess. Why would this be a direct value instead of a guess? The value is set by the variable that defines the input and the function that defines the output, any default here would just be an extra equation causing an overconstrained system. Indeed if someone sets a default then it should be true, but why would it be normal to set an overconstrained system with inputs and outputs? We can make that another kwarg that is nothing by default?

ChrisRackauckas: it should probably just be renamed `guess` and deprecate the old name.

baggepinnen: that would be much less confusing

baggepinnen: I'd prefer if we stop using the word "default" as well in favor of either "start" or "initial", but that's a discussion for another time and place :P

AayushSabharwal: Fixed in https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/291/commits/7b38a539079bad5125367c881de459db46868673

ChrisRackauckas:
Is this failure known?

AayushSabharwal:
It's fixed in https://github.com/SciML/ModelingToolkit.jl/pull/2677

PR #295, opened 2024-05-13
Use `systems = @named begin .. end` idiom to define systems in tutorials

jaakkor2:
Tutorials list all systems in the `ODESystem()`-function call.  This is cumbersome, and one easily forgets to append
the list after adding the actual system and its connections.
```
systems = @named begin
 ..
end
```
is a more user friendly pattern.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,19 +53,20 @@ R = 1.0
 C = 1.0
 V = 1.0
 @variables t
-@named resistor = Resistor(R = R)
-@named capacitor = Capacitor(C = C)
-@named source = Voltage()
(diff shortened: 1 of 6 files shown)


jaakkor2:
Huh,
```
Run julia -e '
┌ Error: Some files have not been formatted !!!
└ @ Main none:6
docs/pages.jl
src/Hydraulic/IsothermalCompressible/components.jl
Error: Process completed with exit code 1.
```
I did not touch `src/Hydraulic/IsothermalCompressible/components.jl` nor `docs/pages.jl`.

PR #296, opened 2024-05-16
Keep non-array connector variable collected

baggepinnen:
The PR
- https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/291

was breaking, in particular, the removal of
```julia
u = collect(u)
```
from `RealInput` and `RealOutput` broke Multibody.jl

Since we have both `RealInput` and `RealInputArray`, where the latter is using symbolic arrays, I propose that we keep the old behavior of `RealInput` intact, i.e., revert the breaking change.

### DIFF
--- a/src/Blocks/utils.jl
+++ b/src/Blocks/utils.jl
@@ -16,8 +16,9 @@
             input = true,
             description = "Inner variable in RealInput $name"
         ]
+        u = collect(u)
     end
-    ODESystem(Equation[], t, [u], []; name = name, guesses = [u => guess])
+    ODESystem(Equation[], t, [u;], []; name = name, guesses = [u => guess])
 end
 @doc """
     RealInput(;name, guess)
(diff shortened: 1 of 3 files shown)


AayushSabharwal:
I think tests might fail here. The reason to not scalarize this is so that the non-scalarized form is available through `guesses`. Once tests pass here, I'll also run MTK tests (using https://github.com/SciML/ModelingToolkit.jl/pull/2676) with this checked out, since it has the potential to break those too

baggepinnen:
> The reason to not scalarize this is so that the non-scalarized form is available through guesses.

Why is this needed, wouldn't using `RealOutputArray` be an option if you need the non-scalarized version?

baggepinnen:
Regarding the test failure here
```julia
  LoadError: Initial condition underdefined. Some are missing from the variable map.
  Please provide a default (`u0`), initialization equation, or guess
  for the following variables:
  
  Any[P₊input₊u(t)]
```
I have tests passing locally on the https://github.com/AayushSabharwal/ModelingToolkit.jl/tree/as/linearization-initialization branch of MTK so this PR will have to wait on 
- https://github.com/SciML/ModelingToolkit.jl/pull/2676

baggepinnen:
tests are passing on MTK master branch

PR #307, opened 2024-07-12
Reduce the extra initial conditions on Translational components

ChrisRackauckas:


ChrisRackauckas:
@ven-k @bradcarman the test cases are ill-formed since they have more equations than unknowns in the initialization, and there seems to be a lot of `= 0` which aren't necessary or expected. I presume these are not how they are defined in the Modelica Standard Library, can we get hits updated this week?

bradcarman:
@ChrisRackauckas  please see https://github.com/bradcarman/ActiveSuspensionModel/blob/bgc/init/ActiveSuspensionModel.jl/test/init.jl

This shows why I think we need to take a couple hours to sort out how this is all supposed to work.  I'm not sure at the moment if I'm running into bugs or if I'm just doing things completely wrong.

ChrisRackauckas:
> # Why am I getting a message that guesses are missing, when they clearly are not...

I don't understand that comment, because the file showed me that there are 0 guesses that are defined.

ven-k:
The JSML/MechanicalComponents doesn't add any initial values. These: https://github.com/JuliaComputing/JuliaSimExampleComponents/blob/main/jsml/active_suspension.jsml#L56-L59 initial values were enough for ActiveSuspension

bradcarman:
Sorry, I didn't explain clearly:  This example https://github.com/bradcarman/ActiveSuspensionModel/blob/bgc/init/ActiveSuspensionModel.jl/test/init.jl redefines all the Standard Library components, all with guesses only and no defaults.  So for testing purposes, this repo doesn't use the MTKStandardLibrary at all.  What we can see with this repo is that there appears to be some issues with MTK currently, removing defaults doesn't seem to work yet.  The Active Suspension model is an attempt to make a complete demo model that requires initialization considering gravity is applied, meaning the springs should have some compression at time 0.  I'm hoping the initialization system can correctly solve for the spring compression.  The problem is setup with zero defaults and only guesses, and then initial equations are specified at the top level.  So far I get only errors.

PR #308, opened 2024-07-12
RFC: Remove zero initialization on Integrator, Derivative, and FirstOrder

ChrisRackauckas:
As started to be discussed in the other thread, these are the only three `= 0` in this package that seem like they have a reasonable argument for it. There is a use case for which the integrator's starting value could be defined by an outside value, in which case this no `= 0` default is better as you will otherwise "accidentally" add an extra initial equation and fail initialization, unless you do `= nothing` (needs to be added).

But, is this better for the normal user?

### DIFF
--- a/src/Blocks/continuous.jl
+++ b/src/Blocks/continuous.jl
@@ -15,12 +15,12 @@ Initial value of integrator state ``x`` can be set with `x`
 
 # Unknowns:
 
-  - `x`: State of Integrator. Defaults to 0.0.
+  - `x`: State of Integrator.
 """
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
@bradcarman @baggepinnen @mtiller-jh for comments.

baggepinnen:
My comments are here https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/306#issuecomment-2217477458 and a few posts after that

ChrisRackauckas:
At least for now, we aren't doing this.

PR #316, opened 2024-08-12
add ability to modify system after analysis point handling

baggepinnen:
add ability to modify system _after_ analysis point handling, but before linearization

This allows conversion to `JuliaSimCompiler.IRSystem` for the structural simplification


ChrisRackauckas:
Needs tests?

PR #324, opened 2024-09-03
step is active at start time

baggepinnen:
for consistency with
- other implementations
- literature
- the other sources in the same file


ChrisRackauckas:
Is this used in DC motor? https://github.com/SciML/ModelingToolkitStandardLibrary.jl/actions/runs/10676494513/job/29590132416?pr=324#step:6:193

baggepinnen:
Yes it is, but that example has been failing for a long time https://github.com/SciML/ModelingToolkitStandardLibrary.jl/actions/workflows/Documentation.yml

PR #334, opened 2024-09-29
Remove defaults from Blocks.SISO

bradcarman:
## Checklist

## Additional context
The SISO component is an extendable component that provides `u` and `y` variables.  These should not be have defaults at the lower level, this makes building components based on them more challenging with the MTK v9 initialization system.


ChrisRackauckas:
Test failure.

We should differentiate the args for guesses from those for the initial values. Looks like this test needs to set an initial.

PR #338, opened 2024-10-18
Update MTK

SebastianM-C:
Updates for the new initialization system in MTK@v9.45.
If the extra u0s are provided, they lead to initialization failures.

## Checklist


ChrisRackauckas (approved):

baggepinnen:
The Translational library uses the velocity connectors that Brad implemented, they require careful initialization of the position of every component since connectors do not carry position information. I would thus guess that the failures are due to some problem with initialization since at least one test is failing a position check.

I would favor us getting rid of this sub-library altogether, but that's a separate issue :)

ChrisRackauckas:
Yeah that test is really bad. It is both overdetermined in the sense of having too many initial conditions, and underdetermined in the sense that it does not even specify the right ones. So everything is fine there, it's just the test is testing for something that isn't necessarily true: it's effectively underdetermined so there are infinitely many initial conditions that satsify the constraints the ones that you choose will change the result. That is a really bad test.

> The Translational library uses the velocity connectors that Brad implemented, they require careful initialization of the position of every component since connectors do not carry position information. I would thus guess that the failures are due to some problem with initialization since at least one test is failing a position check.

> I would favor us getting rid of this sub-library altogether, but that's a separate issue :)

We should just delete it. Indeed this is an issue of not actually specifying position.

PR #358, opened 2025-02-14
Adding Power, Modulo, UnaryMinus, Floor and Ceil blocks 

shobhitvoleti:
## Checklist

## Additional context

A few of the blocks developed for a parallel project being moved to the standard library

### DIFF
--- a/src/Blocks/math.jl
+++ b/src/Blocks/math.jl
@@ -248,7 +248,7 @@ Output the remainder when the first input is divided by second input.
         output = RealOutput()
     end
     @equations begin
-        output.u ~ input1.u % input2.u
+        output.u ~ mod(input1.u, input2.u)
     end
 end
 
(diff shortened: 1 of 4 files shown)


src/Blocks/math.jl
```diff
@@ -208,6 +208,112 @@ Output first input divided by second input.
...
+# Connectors:
+
+  - `input1`
+  - `input2`
+  - `output`
+"""
+@mtkmodel Power begin
+    @components begin
+        input1 = RealInput()
+        input2 = RealInput() # denominator can not be zero
```
baggepinnen, line 225: these could have had better names, like `exponent` instead of `input2`

ChrisRackauckas: True, could you follow up with some suggested names? It hasn't been released yet so it's still fine to change.

baggepinnen: `base, power` or `base, exp`?

shobhitvoleti: Added this PR here that updates names for the Modulo block as well : https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/359

PR #363, opened 2025-03-05
Hydraulic library updated to use macros

matthew-kapp:
## Checklist

- [ ] Appropriate tests were added | No new functionality
- [ ] Any new documentation only uses public API | No new documentation
  
## Additional context

Reviewer @ven-k comments implemented

The models in the Hydraulic library were modified to use @mtkmodel macro instead of @component function. The functionality of the models should remain the same.

Only non-breaking changes were made. Changes include:

- sources.jl
  - MassFlow
  - FixedPressure
  - Pressure
- components.jl
  - Cap
  - Open
  - FlowDivider
  - VolumeBase
  - FixedVolume
  - Volume

The following were NOT changed, as the changes would be breaking:

- components.jl
  - TubeBase
  - Tube
  - ValveBase
  - Valve
  - DynamicVolume
  - SpoolValve
  - SpoolValve2Way
  - Actuator

Additionally, the docstrings were altered to remove P_int and its derivatives, which are no longer used.

The package was tested and only the original fail of VariableResistor occurs.


matthew-kapp:
@ven-k Does this commit interfere/conflict with #367 ?

ven-k:
Yes

matthew-kapp:
@ven-k #367 and your suggestion implemented here additionally

matthew-kapp:
@ven-k ready for review?

src/Hydraulic/IsothermalCompressible/components.jl
```diff
@@ -400,29 +401,29 @@ Fixed fluid volume.
...
+    @components begin
+        port = HydraulicPort()
+    end
 
-    eqs = [D(rho) ~ drho
-           rho ~ full_density(port, p)
-           dm ~ drho * vol]
+    @equations begin
+        dm = port.dm
+        p = port.p
```
ven-k, line 421: This can be added to a
```
begin
...
end
```
(outside `@equations`)

As dm and p are not defined as variables, these equations will throw errors.

src/Hydraulic/IsothermalCompressible/components.jl
```diff
@@ -463,16 +464,14 @@ dm ────►               │  │ area
...
+@mtkmodel Volume begin
 
-        #parameters
-        area,
-        direction = +1, name)
-    pars = @parameters begin
-        area = area
+    @parameters begin
+        area
+        direction = +1
```
ven-k, line 471: `direction` could be of type `@structurtal_parameters`.

Is this intentionally promoted to a `@parameters`?

While here, could you change +1 to 1?

src/Hydraulic/IsothermalCompressible/components.jl
```diff
@@ -482,32 +481,33 @@ See also [`FixedVolume`](@ref), [`DynamicVolume`](@ref)
...
+        f ~ p * area
+        dm ~ drho * x * area + rho * dx * area
+    end
 
-           # physics
-           rho ~ liquid_density(port, p)
-           f ~ p * area
-           dm ~ drho * x * area + rho * dx * area]
 
-    ODESystem(eqs, t, vars, pars; name, systems, defaults = [rho => liquid_density(port)])
```
ven-k, line 506: `defaults = [rho => liquid_density(port)]`

can be added via the `@defaults` block https://docs.sciml.ai/ModelingToolkit/stable/basics/MTKLanguage/#@defaults-begin-block

src/Hydraulic/IsothermalCompressible/components.jl
```diff
@@ -357,36 +357,37 @@ Valve with `area` input and discharge coefficient `Cd` defined by https://en.wik
...
-@component function VolumeBase(; area, dead_volume = 0, Χ1 = 1, Χ2 = 1,
-        name)
-    pars = @parameters begin
+@mtkmodel VolumeBase begin
+
+    @parameters begin
         area = area
         dead_volume = dead_volume
+        x1 = 1
+        x2 = 1
```
ven-k, line 366: x1 x2 were `@structural_parameters`. These could be just `@parameters`. However, that would need fixing these scoping here: https://github.com/SciML/ModelingToolkitStandardLibrary.jl/blob/main/src/Hydraulic/IsothermalCompressible/components.jl#L648-L650 in DynamicVolume

Probably adding them as structural_parameters is easier.

ven-k (commented):
Thanks for the PR.

I have foll. comments:

src/Hydraulic/IsothermalCompressible/components.jl
matthew-kapp: Is this correct?

```julia
    @defaults begin
        rho => liquid_density(port)
    end
```

matthew-kapp:
> Thanks for the PR.
> 
> I have foll. comments:

@ven-k comments implemented. Everything exactly as you said. Only thing I am unsure about is default, which is implemented as follows:

    @defaults begin
        rho => liquid_density(port)
    end

src/Hydraulic/IsothermalCompressible/components.jl
ven-k: Looks like accessing values from subcomponent can't be added to begin...end. My bad.

Alternative fix would be, define dm and p as `@variables`.
And move these back to `@equations` .

matthew-kapp: This seems simplest?

```julia
    @equations begin
        D(rho) ~ drho
        rho ~ full_density(port, port.p)
        port.dm ~ drho * vol
    end
```

ven-k:
Thanks for all the updates.
`@defaults` looks correct.

I see that there are few Hydraulic test failures.
1. It appears `begin...end` doesn't work while accessing values from subcomponents due to scoping issue. My bad. 
Alternative fix would be to promote dm and p as variables. (Without this promoting them to equations will fail) https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/363/files#r1985093021
2. VolumeBase had `X1` and `X2` as args; which is now changed to `x1` and `x2` in some but not all places. That should be fixed. (Changing all mentions to x1, x2 would be preferable.)

matthew-kapp:
@ven-k thanks for the input. I apologize, it's 100% breaking in the tests. I'll need to take my time to fix and debug.

matthew-kapp:
@ven-k back to 572 tests passed

matthew-kapp:
@ven-k Thanks for all the input - much appreciated. I believe it is ready now for review?

ven-k (approved):
LGTM. Thanks.

Formatter failure is related; the test failures are not related.

PR #381, opened 2025-04-14
DOCS: Connect the clock in causal direction

cstjean:
This is really a question maskerading as a PR, but given that `?connect` says:

>   PLEASE NOTE: The connection is assumed to be causal, meaning that
> 
>   `connect(C.output, :plant_input, P.input)`
> 
>   is correct, whereas
> 
>   `connect(P.input, :plant_input, C.output)`
> 
>   typically is not (unless the model is an inverse model).

I'm assuming that `connect(src.input, clk.output)` yields the right answer, but as a casual reader trying to figure out what a `ContinuousClock` is, the causal direction helps understand what's going on.

### DIFF
--- a/docs/src/tutorials/input_component.md
+++ b/docs/src/tutorials/input_component.md
@@ -114,7 +114,7 @@ function MassSpringDamperSystem(data, time; name)
     @named model = MassSpringDamper()
 
     eqs = [connect(model.input, src.output)
-           connect(src.input, clk.output)]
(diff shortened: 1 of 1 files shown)


baggepinnen:
The change is fine. The comment 
```
The connection is assumed to be causal, meaning that
connect(C.output, :plant_input, P.input)
```
only applies to connections that create an analysis point. If there is no analysis point created you can have the arguments to the connect statement in any order

PR #382, opened 2025-04-14
build: bump MTK compat

AayushSabharwal:
## Checklist

- [ ] Appropriate tests were added
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the


AayushSabharwal:
Tests pass locally. It's impossible to be compatible with v9 and v10 because of the change to problem constructors, so we can just tag this. Anyone with v9 won't install it, anyone with v10 is on the bleeding edge anyway and as mentioned earlier tests did pass locally.

Issue #384, opened 2025-04-16
Thermal sensors need outputs


jbatteh:
**Is your feature request related to a problem? Please describe.**

The sensor components in the MTK Thermal library do not have output connectors. Thus they cannot be wired to other components as expected (i.e. PID, etc.).  This request applies to TemperatureSensor, RelativeTemperatureSensor, and HeatFlowSensor.

**Describe the solution you’d like**

The sensed variables should be added to a RealOutput component.

**Describe alternatives you’ve considered**

A clear and concise description of any alternative solutions or features you’ve considered.

**Additional context**

For reference, you can look at Modelica Standard Library (i.e. [Modelica.Thermal.HeatTransfer.Sensors.TemperatureSensor](https://github.com/modelica/ModelicaStandardLibrary/blob/master/Modelica/Thermal/HeatTransfer/Sensors/TemperatureSensor.mo))

bavcol:
When using the sensor components of MTK Electrical, I noticed those are missing the output port as well.
If I find the time, I will create a pull request for those as well.

PR #385, opened 2025-04-24
RealOutputs for thermal sensors

jbatteh:
https://github.com/SciML/ModelingToolkitStandardLibrary.jl/issues/384
Using RealOutputs for thermal sensors.  Note that any existing usage of temperature sensors will need to be updated (i.e. referencing of sensor output goes from T -> T.u).  Connections to sensor output are now possible.   Thermal sensors are now following the same style as those in Rotational with sensor output on connectors.  

Removing usage of OrdinaryDiffEqDefaults in tests

## Checklist

- [ x] Appropriate tests were added
- [x ] Any code changes were done in a way that does not break public API
- [x ] All documentation related to code changes were updated
- [ x] The new code follows the
- [x ] Any new documentation only uses public API
  
## Additional context
No new tests were added but existing tests in Thermal package all pass.  Note that usage of OrdinaryDiffEqDefaults was removed from a few tests.  This package was not included when I instantiated and doesn't seem necessary.  Please check if this change is also OK?

### DIFF
--- a/test/Thermal/piston.jl
+++ b/test/Thermal/piston.jl
@@ -1,4 +1,4 @@
-using ModelingToolkit, OrdinaryDiffEqDefault, Test
+using ModelingToolkit, Test
 using ModelingToolkitStandardLibrary.Thermal
 using ModelingToolkitStandardLibrary.Blocks
 
--- a/src/Thermal/HeatTransfer/sensors.jl
+++ b/src/Thermal/HeatTransfer/sensors.jl
@@ -7,23 +7,18 @@ This is an ideal absolute temperature sensor which returns the temperature of th
 signal. The sensor itself has no thermal interaction with whatever it is connected to. Furthermore, no thermocouple-like
 lags are associated with this sensor model.
 
-# States:
-
-  - `T(t)`: [`K`] Absolute temperature
-
 # Connectors:
 
-  - `port`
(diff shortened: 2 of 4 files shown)


test/Thermal/piston.jl
```diff
@@ -1,4 +1,4 @@
-using ModelingToolkit, OrdinaryDiffEqDefault, Test
```
ChrisRackauckas, line 1: remove from the test part of the Project toml as well?

ChrisRackauckas:
@baggepinnen does this constitute a breaking change?

baggepinnen:
Technically yes, any model that previously used this blocks will be broken. The blocks were pretty useless since they had no output connector, so in practice I'm not sure if they were used all that much and thus if this PR will actually break anything

test/Thermal/piston.jl
```diff
@@ -1,4 +1,4 @@
...
+using ModelingToolkit, Test
```
ChrisRackauckas, line 1: what's the point of the removal if OrdinaryDiffEq is directly used in others?

ChrisRackauckas:
Okay so @baggepinnen you agree with these changes?

test/Thermal/piston.jl
baggepinnen: I use TestEnv.jl yoninstantiate the test environment, which is generally different from the package environment, notably, tests often require additional dependencies.

PR #392, opened 2025-05-12
Fix initialization for TranslationalPosition and add more tests

SebastianM-C:
## Checklist

## Additional context

The initializations produced by the `TranslationalPosition` components are very overconstrained. I think that there were some initialization workarounds that were needed in the past, but on more modern MTK versions these should no longer be needed.

One  of the things that I removed is the  equations that were`flange.s` from higler level components shuch as the `Spring` and `Damper`. Is this breaking?


baggepinnen:
I have no objections, the `Translational` and `TranslationalPosition` have had so many warts historically that I found them unusable and added the `TranslationalModelica` library for use from Multibody.jl instead. 

In the future, I say we remove both of those and use only `TranslationalModelica` (renamed to `Translational`)

PR #417, opened 2025-12-04
Replace custom clamp function with Base.clamp

baggepinnen:
I think that `Base.clamp` should work nowadays. This would make it easier to spot clamp operations in equations (rather than `max(min(...))` and replace them during linearization.


ChrisRackauckas (approved):

baggepinnen:
It is
```julia
julia> @variables x
1-element Vector{Num}:
 x

julia> clamp(x, 0, 1)
clamp(x, 0, 1)
```

PR #450, opened 2026-05-01
test: declare dangling sensor / capacitor variables as outputs

ChrisRackauckas-Claude:
## Summary

After the v7 import fixes (#445) and docs fixes (#448, #449), several **Tests** jobs hit `ExtraVariablesSystemException` from MTK 11's stricter `mtkcompile` balance check. The dangling variables aren't bugs in the test models — they're sensor / output connectors the test never feeds back, so semantically they're **outputs**. Declare them via `outputs=` so `mtkcompile` excludes them from the balance check. **No `fully_determined=false`. No skipped tests.**

- `test/Blocks/utils.jl` "SISO Check" → `outputs = [so.output.u]`.
- `test/Thermal/thermal.jl` "Heat flow system" → `outputs = [mass1.T, hf_sensor2.port_b.Q_flow]`.
- `test/Mechanical/rotational.jl` "sensors" → switched from `@mtkcompile` to the explicit form `@named model = Sensors(); sys = mtkcompile(model; outputs = [...])`, listing the 11 dangling sensor / connector variables.
- `test/Mechanical/translational.jl` "AccelerationSensor" → `outputs = [mass.v, mass.flange.s]`.

## Out of scope (intentionally still failing — neither disabled nor skipped)

- **`test/Blocks/sources.jl` BSplineInterpolation / Initialization** → `mtkcompile` of a `ParametrizedInterpolation` system fails inside SymbolicUtils' `promote_symtype` with `MethodError: fntype_X_Y(::Type{Real})`. The only existing `fntype_X_Y` method takes `Type{<:SymbolicUtils.FnType{X, Y}}`, so something is calling `promote_symtype` with a bare `Real` symtype. Root cause is in either SymbolicUtils itself or the `(interpolator::interp_type)(..)::eltype(u)` declaration in `src/Blocks/sources.jl:903–904` (the callable parameter is ending up with `symtype == Real` instead of a function type).
- **`test/Blocks/test_analysis_points.jl`** `ModelingToolkit.get_looptransfer` call → MTK initialization throws `"Cyclic guesses detected in the system"` from `varmap_to_vars` while building the initialization problem (`(P.xˍt(t))[2] => (P.xˍt(t))[2]`). Needs an MTK fix or an explicit `missing_guess_value` keyword.

## Test plan

- [ ] `Tests (1, Core)`, `Tests (lts, Core)`, `Tests (pre, Core)` no longer raise `ExtraVariablesSystemException` from the four sites above.
- [ ] The two upstream-blocked tests above still fail and the failures point at SymbolicUtils / MTK respectively.

### DIFF
--- a/test/Blocks/utils.jl
+++ b/test/Blocks/utils.jl
@@ -32,7 +32,7 @@ end
     @named c = Constant(; k = 1)
     @named so = SecondOrder(; k = k, w = w, d = d, xd = 1)
     @named iosys = System(connect(c.output, so.input), t, systems = [so, c])
-    sys = mtkcompile(iosys)
+    sys = mtkcompile(iosys; outputs = [so.output.u])
 
     initsys = ModelingToolkit.generate_initializesystem(sys)
     initsys = mtkcompile(initsys)
--- a/test/Mechanical/rotational.jl
+++ b/test/Mechanical/rotational.jl
@@ -278,7 +278,22 @@ end
         end
     end
 
-    @mtkcompile sys = Sensors()
+    @named model = Sensors()
+    sys = mtkcompile(model;
+        outputs = [
+            model.inertia1.w,
+            model.damper.phi_rel,
+            model.damper.w_rel,
+            model.damper.a_rel,
+            model.inertia2.phi,
+            model.inertia2.w,
+            model.rel_speed_sensor.phi_rel,
+            model.rel_speed_sensor.w_rel.u,
+            model.fixed.flange.phi,
+            model.fixed.flange.tau,
+            model.torque_sensor.flange_a.tau,
+        ]
+    )
 
     prob = ODEProblem(sys, [D(D(sys.inertia2.phi)) => 0.0], (0, 10.0))
     sol = solve(prob, Rodas4())
--- a/test/Mechanical/translational.jl
+++ b/test/Mechanical/translational.jl
@@ -248,8 +248,13 @@ end
             @named sys = System(
                 eqs, t, [], []; systems = [force, source, mass, acc, acc_output]
             )
-            s = complete(mtkcompile(sys))
-            prob = ODEProblem(s, [], (0.0, pi), fully_determined = true)
+            s = complete(mtkcompile(sys;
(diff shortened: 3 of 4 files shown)


baggepinnen:
@AayushSabharwal should declaring outputs ever be needed?

ChrisRackauckas-Claude:
Closing — local testing showed this PR does **not** actually fix the `ExtraVariablesSystemException` failures it claims to fix.

Concrete result from running `test/Blocks/utils.jl` "SISO Check" locally with this PR's `mtkcompile(iosys; outputs = [so.output.u])`:

```
SISO Check: Error During Test at test/Blocks/utils.jl:30
  ExtraVariablesSystemException: The system is unbalanced. There are 4 highest order derivative variables and 3 equations.
  More variables than equations, here are the potential extra variable(s):
   so₊xdˍt(t)
```

Same for "Heat flow system" — fails on `mass1₊port₊T(t)` / `hf_sensor2₊port_a₊Q_flow(t)`.

Why `outputs=` doesn't help: looking at MTK's `_mtkcompile!` (`systems/systemstructure.jl:240`), the `outputs` kwarg is just stored as `sys.outputs = outputsyms`. The balance check (`StateSelection.check_consistency` in `StateSelection/utils.jl:64–71`) does `n_highest_vars == neqs` and only `inputs` (via `inputs_to_parameters!`) reduce `n_highest_vars`. Outputs are never excluded from the count.

The probe-equation alternative (`@variables y_probe(t); y_probe ~ so.output.u`) doesn't help either — it adds 1 var and 1 equation, so the imbalance just shifts to a different dangling variable.

Real next steps that don't disable tests:
1. File an MTK regression issue: `mtkcompile` on a `SecondOrder + Constant` system used to compile and now reports `ExtraVariablesSystemException` even though the equations clearly balance through alias elimination.
2. Same for the heat-flow / rotational-sensors / translational-AccelerationSensor models — connector flow/across variables that should be eliminated by `expand_connections` are being flagged as dangling.

Both are upstream issues, not anything fixable inside this repo without changing what the tests actually test.

PR #477, opened 2026-07-12
test: use Base.ifelse in Chua circuit

ChrisRackauckas-Claude:
## Summary

Remove the stale `using IfElse: ifelse` from the Chua circuit test. The test now uses Julia's built-in `Base.ifelse`, matching the production-code migration in #473.

## Root cause

Commit `dca3a237b1cda8c6befdf5bd0ef53f4c03fbd9b9` removed IfElse.jl from the package dependencies and migrated production call sites, but missed `test/chua_circuit.jl`. The Core group has therefore failed at test loading since #473; the first bad release is v2.29.3 (v2.29.2 is the last release before the removal), and v2.29.4 remains affected.

This also breaks ModelingToolkit's downstream Core job on both master and unrelated PRs:

- [ModelingToolkit master](https://github.com/SciML/ModelingToolkit.jl/actions/runs/29186963983/job/86634738179)
- [ModelingToolkit PR #4730](https://github.com/SciML/ModelingToolkit.jl/actions/runs/29190999984/job/86645537310)
- [First failing Core job on #473](https://github.com/SciML/ModelingToolkitStandardLibrary.jl/actions/runs/28731796831/job/85201145714)

## Verification

Using Julia 1.12.6, ModelingToolkit `ee7dd34d8b7cfe8a085e0fa04cec28a607053f66`, and the reusable downstream workflow locally:

```text
# Unmodified v2.29.3
Core/analog.jl       | Pass 99  Total 99
Core/chua_circuit.jl | Error 1  Total 1
ArgumentError: Package IfElse not found in current path.

# Current main plus this patch
Core/analog.jl       | Pass 99  Total 99
Core/chua_circuit.jl | Pass 1   Total 1
```

`Runic v1.7.0 --check .` also completed successfully.

The full Core run continued past Chua and exposed three unrelated `isothermal_compressible.jl` `ExtraVariablesSystemException` errors. Those are outside this one-line test-import fix and are being routed to a separate clean-main reproduction/bisect; they were previously masked by the earlier Chua load error.

**Please ignore this PR until reviewed by @ChrisRackauckas.**

Issue #457, opened 2026-06-06
Test suite red on master (latest MTK/SymbolicUtils): _clamp import + ExtraVariablesSystemException + fntype_X_Y; blocks downgrade re-enable


ChrisRackauckas-Claude:
Investigated re-enabling downgrade CI. The suite fails on **master with the latest deps** (MTK 11.26.8, MTKBase 1.42.2, SciMLBase 3.18.0, DiffEqBase 7.5.5, OrdinaryDiffEq 7.0.0, SymbolicUtils 4.34.3) — so these are pre-existing master failures, not downgrade-floor problems:

1. **`_clamp` import bug (clear):** `test/Blocks/math.jl:4` does `using ModelingToolkitStandardLibrary.Blocks: _clamp, _dead_zone`, but `_clamp` is defined nowhere in `src/` (HEAD df33dee — `src/Blocks/nonlinear.jl` defines only `_dead_zone`; `Limiter` uses built-in `clamp`). → `UndefVarError: _clamp`.
2. **`ExtraVariablesSystemException` (deeper):** `SISO Check`, `Heat flow`, Mechanical `sensors`, `AccelerationSensor` build unbalanced systems out of `mtkcompile` under current MTK.
3. **`fntype_X_Y` MethodError:** `LinearInterpolation`/`BSplineInterpolation` hit `MethodError: no method matching fntype_X_Y(::Type{Real})` under SymbolicUtils 4.34.3.

(2) and (3) look like MTK/SymbolicUtils co-evolution breakage and reproduce at newest versions. Fixing (1) alone won't green the suite.

For when master is fixed, the **resolve/precompile floors** that work on Julia 1.10 (verified): ModelingToolkit ≥11.2.0 (MTK 11.0.0/11.1.0 call `Base.ispublic`, a Julia-1.11-only function, in `@import_mtkbase` → `UndefVarError: ispublic` on 1.10; fixed in v11.2.0), and a DiffEqBase/SciMLBase pairing that avoids both the `init` kwcall ambiguity (DiffEqBase <6.190.0) and the `numargs` rrule method-overwrite (DiffEqBase ≥6.210.1 with SciMLBase ≥2.146.0, or keep SciMLBase <2.146.0).

Downgrade CI can't be re-enabled until the master suite is green. (Possibly related to the OrdinaryDiffEq 7.0.0 issues #3721/#3722.)

_From the SciML downgrade re-enablement pass._

PR #478, opened 2026-07-12
Fix isothermal domain test connections

ChrisRackauckas-Claude:

## Summary

- Use `domain_connect` for hydraulic domain propagation instead of routing flow equations through the fluid parameter setter.
- Keep the cap/volume physical connection explicit.
- Transfer the Actuator operating point with public symbolic problem indexing instead of pairing symbolic unknowns with raw solution storage.
- Check domain parameter propagation through the public `ModelingToolkitBase.bindings` API.
- Add source/volume flow-balance assertions to the DynamicVolume test.

## Root cause

ModelingToolkit commit `b0f315369` stopped silently zeroing underconstrained variables during alias elimination. The immediate parent `80f7f62bc` passes the reduced DynamicVolume model, while `b0f315369` reports 48 highest-order variables and 47 equations. The tests were using a fluid domain setter as a physical flow connector, which hid the model structure behind the removed zeroing behavior.

The Actuator test also copied a failed zero-length solution by zipping `unknowns(initsys)` with raw `initsol.u` storage. Its public problem operating point contains `piston.dx = 0`, while the copied raw value was `NaN`. Reading each unknown through `initprob[v]` preserves the intended operating point.

## Validation

- Rebased head `1cb83378` on current `main` `8539d8f2` (merged #477).
- Direct `test/isothermal_compressible.jl` on Julia 1.12.6, exit 0:
  - Fluid Domain and Tube: 2/2
  - Valve: 1/1
  - DynamicVolume and minimum-volume feature: 12/12
  - Actuator System: 9/9
  - Prevent Negative Pressure: 3/3
- Exact commit boundary with the same reduced model and local environment:
  - `80f7f62bc`: exit 0
  - `b0f315369`: exit 1, 48 variables / 47 equations
- Official `GROUP=Core Pkg.test()` on a temporary validation stack containing current main, this patch, and #480: Analog 100/100, Chua 1/1, Continuous 76/76, Digital 34/34, and Isothermal 27/27; the run then stopped on the separate clean-main exact-equality assertion in `test/magnetic.jl:66`, which is under an independent reproduction and ULP analysis.
- Repository-wide Runic 1.7 `--check .`: exit 0 across 81 files.
- `git diff --check`: exit 0.

No test was skipped, disabled, silenced, or made warn-only.

Related: #457, merged #477, and #480.

### DIFF
--- a/test/isothermal_compressible.jl
+++ b/test/isothermal_compressible.jl
@@ -1,4 +1,6 @@
 using ModelingToolkit, OrdinaryDiffEq, Test
+using ModelingToolkitBase: bindings, domain_connect, unknowns
+using SciMLBase: NoInit
 using SciCompDSL
 using OrdinaryDiffEqNonlinearSolve: NLNewton
 using ModelingToolkit: t_nounits as t, D_nounits as D
@@ -152,8 +154,7 @@ end
         end
 
         eqs = [
-            connect(fluid, src1.port)
-            connect(fluid, src2.port)
+            domain_connect(fluid, src1.port, src2.port)
             connect(src1.port, vol1.port)
             connect(src2.port, vol2.port)
             connect(vol1.flange, mass.flange, vol2.flange)
@@ -196,6 +197,8 @@ end
     # volume/mass should stop moving at opposite ends
     @test sol(0; idxs = sys.vol1.x) == 0.1
     @test sol(0; idxs = sys.vol2.x) == 0.1
+    @test sol[sys.src1.port.dm] ≈ -sol[sys.vol1.port.dm]
+    @test sol[sys.src2.port.dm] ≈ -sol[sys.vol2.port.dm]
 
     @test round(sol(1; idxs = sys.vol1.x); digits = 2) == 0.19
     @test round(sol(1; idxs = sys.vol2.x); digits = 2) == 0.01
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Rebase and validation update:

- Upstream `main` advanced to `8539d8f2` when #477 merged. I stopped before pushing, fetched both remotes, and verified the feature remote still contained only the expected old head `7b7966a9`.
- Rebased the single isothermal commit onto that new main; new head is `1cb83378` and the required co-author trailer is preserved.
- Re-ran the complete focused file on Julia 1.12.6: 27/27 assertions passed with `isothermal_rebase_test_exit=0`.
- Re-ran Runic across all 81 repository files and `git diff --check`: both exited 0.
- Updated the remote with an exact-SHA `--force-with-lease`; CI has restarted.

The official integration Core run now reaches and passes Isothermal 27/27, then exposes only the separately audited magnetic one-ULP assertion.

ChrisRackauckas-Claude:
Post-rebase native `GROUP=Core Pkg.test()` evidence on Julia 1.12.6 (head `1cb83378`):

- `Core/analog.jl`: 100/100
- `Core/chua_circuit.jl`: 1/1
- `Core/continuous.jl`: 76/76
- `Core/demo.jl`: 1/1
- `Core/digital.jl`: 34/34
- `Core/isothermal_compressible.jl`: 27/27

The group then stopped at `Core/magnetic.jl` (2 pass, 1 exact-equality failure). That same one-ULP-scale flux residual reproduces on unmodified current `main` and is being handled in a separate clean-main audit; it is not introduced by this PR.

PR #479, opened 2026-07-12
Raise SymbolicUtils floor for minimum dependency tests

ChrisRackauckas-Claude:
Ignore this PR until it has been reviewed by @ChrisRackauckas.

## Summary

- raise the SymbolicUtils compatibility floor to 4.38.1
- assert that the HeatingDiode simulation terminates successfully before inspecting its solution

## Root cause

The minimum-dependency workflow started selecting SymbolicUtils 4.10.0 after SymbolicUtils became a direct dependency. With the workflow's ModelingToolkit 11.2.0 / OrdinaryDiffEq 6.103.0 stack, SymbolicUtils 4.10.0 leaves a `-3.48e-14` algebraic residual at the reverse-current bound. The computed minimum current is consequently `-1.0000000347787363e-6`, just below the model's existing physical assertion of `-1e-6`.

The same diagnostic stack produces exactly `-1e-6` with SymbolicUtils 4.38.1 and 4.40.0. Tightening the solver tolerances did not remove the sign-sensitive residual with 4.10.0, so this keeps the test assertion unchanged and raises the dependency floor to the first verified good release instead.

## Validation

- `Runic 1.7.0 --check Project.toml test/analog.jl`: passed
- exact Julia 1.10 minimum-dependency workflow with SymbolicUtils 4.38.1: `Core/analog.jl | 100 pass, 100 total`
- the full test command proceeded past analog and then encountered the unrelated clean-main `IfElse` test-dependency error in `Core/chua_circuit.jl`; that blocker is already addressed by draft PR #477

Failing CI reproduced from https://github.com/SciML/ModelingToolkitStandardLibrary.jl/actions/runs/29193734380/job/86652873871.


ChrisRackauckas-Claude:
CI checkpoint:

- Exact local downgraded Core/analog.jl passed 100/100 with SymbolicUtils 4.38.1.
- CI independently passes Core/analog.jl 100/100 on Julia current, LTS, prerelease, and the downgrade job.
- Each Core/downgrade job then stops at the unrelated clean-main Chua test dependency error, which is addressed by draft PR #477.
- QA, Runic, suggestions, spelling, and matrix detection are green.
- The docs job reaches separate clean-main custom_component.md and input_component.md example failures; those are being reproduced independently without weakening Documenter error handling.

No assertion was loosened or skipped. The HeatingDiode successful-retcode regression and existing strict current bound both pass.

ChrisRackauckas-Claude:
Post-merge downstream CI classification:

- `ModelingToolkit.jl/All` completed after 2h23m47s.
- Earlier MTK groups passed, including `InterfaceII` (945 pass / 7 existing broken) and `SymbolicIndexingInterface` (1964/1964).
- The terminal `Downstream` group errored only when its environment resolved LinearSolve 4.3.0 after the parent test process had already loaded LinearSolve 3.87.0. `LinearSolveArnoldiMethodExt` then tried to access `LinearSolve.ArnoldiMethodJL` on the already-loaded older module; the log explicitly reports `LinearSolve v4.3.0 [loaded: v3.87.0]`.
- This is the exact mixed-version clean-master condition previously reproduced and fixed in closed ModelingToolkit PR #4730. Per the prior closure, I am not reopening or duplicating that PR.

This failure occurs after the SymbolicUtils-floor change's relevant MTK groups passed and is not caused by this PR.

PR #490, opened 2026-07-13
Initialize translational acceleration sensor mass

ChrisRackauckas-Claude:
This PR should be ignored until reviewed by @ChrisRackauckas.

## Summary

- initialize the TranslationalPosition acceleration-sensor fixture's mass position and velocity at component construction
- keep the fixture fully determined under current ModelingToolkit initialization checks

## Root cause

`TP.Mass` stopped supplying `s = 0` and `v = 0` defaults in #305, but this fixture continued constructing it with only `m`. ModelingToolkit v11.10.0 exposed that underconstraint after `b0f31536949370d189c03df71bc38f692a9a6975` stopped alias elimination from silently zeroing underconstrained variables. With ModelingToolkit v11.31.2, `ODEProblem(...; fully_determined = true)` therefore reports a structurally singular `mass₊v(t)` initialization system.

Providing the initial state on `TP.Mass` is the smallest fix and establishes the state before `mtkcompile`. This extracts the same focused fixture correction from the broader #452; the `outputs` approach attempted in #450 does not alter structural balance.

## Local verification

- Julia 1.10.11: `test/translational.jl` — 16/16 passed
- Julia 1.12.6: `test/translational.jl` — 16/16 passed after rebasing onto `06464384`
- Julia 1.13.0-rc1: `test/translational.jl` — 16/16 passed
- explicit acceleration diagnostic — maximum error against `sin(4πt) / 4` was `0.0`
- Runic v1.7.0: full-repository `--check` exited 0
- native stacked `GROUP=Core` with #485–#488: `Core/translational.jl` passed 16/16; the overall group later exited 1 in the unrelated `Core/utils.jl` SISO check, which is being handled as a separate clean-main audit

PR #491, opened 2026-07-13
test: exercise SISO initialization through ODEProblem

ChrisRackauckas-Claude:
**Please ignore this PR until reviewed by @ChrisRackauckas.**

## Summary

- Construct the SISO test's initialized system through `ODEProblem` instead of solving the standalone `InitializationProblem`.
- Keep both behavioral assertions exact and unchanged at `1.0`.
- Import `ODEProblem` from its declaring public module, `SciMLBase`, and remove the now-unused `OrdinaryDiffEq` import.

This is a focused extraction of the SISO change from draft PR #452. That PR contains unrelated changes and conflicts; none of those changes are included here. The branch is now rebased on current `main` after #489 merged, so this PR contains one commit and changes only `test/utils.jl`.

## Root cause

PR #469 replaced the old manual `generate_initializesystem` / `NonlinearProblem` path with `InitializationProblem`. A standalone initialization problem is keyed to its intermediate initialization system, not the fully initialized original ODE system. This has two resolver-dependent failure modes:

- Older minimum-compatible stack: `initsol[sys.so.xd]` is `0.0`, not the declared `1.0` initial value.
- Current downgrade stack: `sys.so.u` has been eliminated from the initialization system, so indexing it throws `ArgumentError: Symbol so₊u(t) is not present in the system`.

`ODEProblem` performs the complete operating-point / initialization preprocessing and retains the public indexing semantics for both original states and observed variables. The test still checks exactly the same two values; it now checks them on the object whose initialization behavior matters to users.

A signature-specific source bisect identifies `3672b875` (PR #469) as the first commit producing the missing-observed-symbol behavior; its parent `a79a77f3` passes this SISO check on the preserved graph. `git log --follow` confirms #469 is still the latest `main` edit to this test.

## Exact current-main reproduction

The current-main downgrade run [29870067371](https://github.com/SciML/ModelingToolkitStandardLibrary.jl/actions/runs/29870067371), [job 88767752470](https://github.com/SciML/ModelingToolkitStandardLibrary.jl/actions/runs/29870067371/job/88767752470), fails at `test/utils.jl:41` with the missing-`so₊u` error: 13 pass / 1 error in `Core/utils.jl`.

[...]

PR #492, opened 2026-07-13
test: check SISO initialization via ODEProblem

ChrisRackauckas-Claude:
**Please ignore this PR until it has been reviewed by @ChrisRackauckas.**

## Summary

- exercise the SISO initialization assertions through `SciMLBase.ODEProblem`
- keep both exact checks (`xd == 1.0` and `u == 1.0`) unchanged
- remove the test-only `OrdinaryDiffEq` import because this test no longer solves a standalone initialization problem

This is a fresh, focused extraction of the SISO test-path correction discussed in the stale omnibus #452. It does not include any of #452's unrelated changes.

## Why

PR #469 changed this test from the earlier explicit `generate_initializesystem`/`mtkcompile`/`NonlinearProblem` flow to `InitializationProblem(sys, 0.0)`, then indexed that standalone problem's solution with symbols from the original ODE system.

A standalone `InitializationProblem` represents the generated initialization system. Its symbolic container is not a contract that every original ODE unknown or observed alias remains indexable, and its intermediate unknown/default representation varies across compatible dependency graphs:

- current graph: `xd` initializes to `1.0`, while indexing `so.u` errors because that alias is not present in the generated initialization system
- preserved minimum graph (`ModelingToolkit 11.0.0`, `ModelingToolkitBase 1.1.0`, `Symbolics 7.4.1`, `SciMLBase 2.125.0`): the standalone path returns `xd == 0.0` and `u == 1.0`
- on that same minimum graph, the pre-#469 explicit path returns `xd == 1.0` and `u == 1.0`

The public `ODEProblem` construction is the operation whose contract the test actually wants: it runs initialization for the ODE and retains the original system as the symbolic provider. Both original assertions therefore remain exact and continue to test the intended initialized values.

## Local validation

- Julia 1.10.11, current resolved graph, complete `test/utils.jl`: 8/8 passed
- Julia 1.12.6, current resolved graph, complete `test/utils.jl`: 8/8 passed
- Julia 1.13.0-rc1 with `--compiled-modules=no`, current graph, complete `test/utils.jl`: 8/8 passed (source mode was needed because the local stdlib JLL cache had incompatible precompile flags)
- preserved exact minimum graph, focused public `ODEProblem` path: `xd == 1.0`, `u == 1.0`
- whole-repository `Runic --check .`: exit 0
- `git diff --check`: exit 0

`GROUP=Core julia +1.12 --project=. -e 'using Pkg; Pkg.test()'` ran all preceding Core files successfully, then exited 1 at the known clean-main `AccelerationSensor` error in `test/translational.jl:235` (`InvalidSystemException`, structurally singular `mass₊v(t)`). That unrelated base failure is addressed separately by #490; this PR does not alter or include that fix.

## Investigation steps

1. Reproduced both reported SISO symptoms on an unmodified current checkout and a preserved exact minimum manifest.
2. Inspected the compiled ODE system and generated initialization systems to identify the symbol-container mismatch.
3. Re-ran the pre-#469 path on the exact minimum graph to locate the source boundary.
4. Validated the public `ODEProblem` path on current Julia 1.10/1.12/1.13 and the exact minimum dependency graph.
5. Ran the repository-native Core harness, whole-repository Runic, and whitespace checks.

### DIFF
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -1,7 +1,7 @@
 using ModelingToolkitStandardLibrary.Blocks
 using SciCompDSL
 using ModelingToolkit
-using OrdinaryDiffEq
+using SciMLBase: ODEProblem
 using ModelingToolkit: t_nounits as t, D_nounits as D
 using Symbolics
 
@@ -34,11 +34,10 @@ end
     @named iosys = System(connect(c.output, so.input), t, systems = [so, c])
     sys = mtkcompile(iosys)
 
-    initprob = ModelingToolkit.InitializationProblem(sys, 0.0)
-    initsol = solve(initprob)
+    prob = ODEProblem(sys, [], (0.0, 1.0))
 
-    @test initsol[sys.so.xd] == 1.0
-    @test initsol[sys.so.u] == 1.0
+    @test prob[sys.so.xd] == 1.0
+    @test prob[sys.so.u] == 1.0
 end
 
 @test_deprecated RealInput(; name = :a, u_start = 1.0)


ChrisRackauckas-Claude:
Closing this draft as a duplicate of #491, which already carries the focused SISO correction and its validation. No changes from this branch need to be merged separately.

PR #494, opened 2026-07-17
Run downgrade CI on pushes to main

ChrisRackauckas-Claude:
## Summary

- run the `Downgrade` workflow after pushes to the repository's actual default branch, `main`
- leave pull-request filtering, path filtering, the reusable workflow, and the downgrade configuration unchanged

## Root cause

The repository moved from `master` to `main`, but the historical branch-reference update changed only the `pull_request` target in `Downgrade.yml`. The `push` target remained `master`, so merges to the default branch did not trigger the downgrade workflow.

## Impact

Merged default-branch changes will again receive strict minimum-version coverage. This is a workflow-trigger-only change; it does not alter package behavior or dependency bounds.

## Validation

- `actionlint -verbose .github/workflows/Downgrade.yml` — exit 0, zero parse/rule errors
- `julia +1.12 --startup-file=no --project=../../audit/runic_env -m Runic --check --diff .` — exit 0, no formatting diff
- shell assertion over the `push` block — exactly one `main` target and no `master` target
- `git diff --check` — exit 0

Package tests were not run because the only changed datum is a GitHub Actions event filter; the package source, tests, dependency bounds, and reusable downgrade job are untouched.

## Known independent failure

An exact version-only run on clean `main` can expose the existing SISO initialization failure. Draft #491 addresses that package behavior separately. This PR neither hides that failure nor duplicates its fix.

## Process

- fetched the canonical repository and verified the authenticated fork's `main` is at the same commit (`06464384`)
- inspected the workflow's full history and the prior canonical-CI PRs before editing
- made the one-line change on a feature branch and pushed it without force

Ignore this PR until it has been reviewed by @ChrisRackauckas.

### DIFF
--- a/.github/workflows/Downgrade.yml
+++ b/.github/workflows/Downgrade.yml
@@ -7,7 +7,7 @@ on:
       - 'docs/**'
   push:
     branches:
-      - master
+      - main
     paths-ignore:
       - 'docs/**'
 jobs:


ChrisRackauckas-Claude:
Implementation scratchpad for commit `a051ff76`:

1. Confirmed `main` is the repository default and compared the other push-triggered workflows, which already target `main`.
2. Traced `Downgrade.yml` history: `bc036599` updated the pull-request branch from `master` to `main` but left the push branch unchanged.
3. Changed only that stale push branch; no package/test/compat or reusable-workflow settings changed.
4. Ran actionlint (zero errors), Runic (no diff), an exact trigger assertion (`main` once, `master` absent), and `git diff --check` (clean).
5. Did not run package tests because this event-filter edit cannot affect package execution. The known clean-main SISO failure remains isolated to #491 and was not skipped, silenced, or duplicated here.

ChrisRackauckas-Claude:
CI classification after the first PR run:

- `Downgrade Tests` was triggered and reached the package suite, which verifies the branch-filter objective of this PR. It stopped at the independently known `SISO Check` failure in `test/utils.jl` (13 pass, 1 fail); draft #491 addresses that clean-`main` behavior and is intentionally separate.
- Core on Julia current, LTS, and prerelease all stopped at the same `AccelerationSensor` structurally-singular initialization fixture in `test/translational.jl`. That clean-`main` regression was independently reproduced, bisected, and fixed in draft #489; #491 is stacked on that fix.
- The downstream ModelingToolkit job is also unrelated to this workflow-only diff. Its log first loads `LinearSolve v3.87.0`, later resolves `LinearSolve v4.3.0 [loaded: v3.87.0]`, and then attempts to precompile the v4 `LinearSolveArnoldiMethodExt` against the already-loaded v3 module, producing `UndefVarError: ArnoldiMethodJL not defined in LinearSolve`. The same signature is present on the unrelated #489 and #491 branches.
- Documentation, QA, Runic, Runic suggestions, and spell checking passed.

No additional change belongs in this trigger-only PR: its new `main` push trigger executed as intended, and it neither causes nor masks the package/base failures above.

